PR #2, opened 2022-02-03
Upgradeability for SubscriptionManager

cygnusv:
* Introduces upgradeability for the SubscriptionManager contract using OpenZeppelin upgradeability framework
* Changes the policy cost model to include the number of nodes in the policy (referred to as `size`). New cost formula is: `cost = feeRate * duration * size`, where `feeRate` is denominated in Matic wei per second.
* Defines an initial fee rate of **1 matic gwei per second**
* Adds some management roles to SubscriptionManager:
    * WITHDRAW_ROLE: Power to withdraw funds from SubscriptionManager
    * SET_RATE_ROLE: Power to set the fee rate

### DIFF
--- a/scripts/deploy_subscription_manager.py
+++ b/scripts/deploy_subscription_manager.py
@@ -9,8 +9,8 @@ from pathlib import Path
 
 INITIAL_FEE_RATE = Wei("1 gwei")
 
-def main():
-    deployer = accounts.load('test')
+def main(account_name : str = 'test'):
+    deployer = accounts.load(account_name)
 
     # TODO: There has to be a better way to do this resolution. See #13
     current_project = project.get_loaded_projects().pop()


contracts/matic/SubscriptionManager.sol
```diff
@@ -0,0 +1,148 @@
...
+    function getPolicy(bytes16 _policyID) public view returns(Policy memory){
+        return _policies[_policyID];
+    }
+
+    function isPolicyActive(bytes16 _policyID) public view returns(bool){
+        return _policies[_policyID].endTimestamp > block.timestamp;
+    }
+
+    function _setFeeRate(uint256 newFee) internal {
+        uint256 oldFee = feeRate;
```
derekpierre, line 133: Do we need a `require` here to ensure that newFee is always > 0?

contracts/matic/SubscriptionManager.sol
```diff
@@ -0,0 +1,148 @@
...
+    struct Policy {
+        address payable sponsor;
+        uint32 startTimestamp;
+        uint32 endTimestamp;
+        uint16 size; // also known as `N`
+        // There's still 2 bytes available here
+        address owner;
+    }
+
+    event PolicyCreated(
```
derekpierre, line 26: If we envision the feeRate possibly being changed, then should we include its value in the event?

scripts/deploy.py
```diff
@@ -1,7 +1,28 @@
 #!/usr/bin/python3
```
derekpierre, line 1: should this file be named specific to subscription manager eg. `deploy_subscription_manager.py` OR do we envision this single script being more customizable once other contracts are deployed from this repos.

contracts/matic/SubscriptionManager.sol
```diff
@@ -0,0 +1,148 @@
...
+        address indexed sponsor,
+        address indexed owner,
+        uint16 size,
+        uint32 startTimestamp,
+        uint32 endTimestamp
+    );
+
+    event FeeRateUpdated(uint256 oldFeeRate, uint256 newFeeRate);
+
+    // Per-second service fee rate
```
vzotova, line 37: maybe also add "per node"?

contracts/matic/SubscriptionManager.sol
cygnusv: Not necessarily. A fee rate of 0 is plausible.

contracts/matic/SubscriptionManager.sol
cygnusv: That's a good point. Maybe rather than the value of the feeRate, we should add here the total cost for the Policy. What do you think?

scripts/deploy.py
cygnusv: You're totally right.

contracts/matic/SubscriptionManager.sol
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    function _setFeeRate(uint256 newFee) internal {
+        uint256 oldFee = feeRate;
+        feeRate = newFee;
+        emit FeeRateUpdated(oldFee, newFee);
+    }
+
+    function setFeeRate(uint256 _rate_per_second) onlyRole(SET_RATE_ROLE) external {
+        _setFeeRate(_rate_per_second);
```
cygnusv, line 139:
```suggestion
        _setFeeRate(_ratePerSecond);
```

contracts/matic/SubscriptionManager.sol
derekpierre: Total cost of the policy sounds fine to me. 

@vzotova may have some thoughts here - the old PolicyCreated event had `feeRate` in it.

cygnusv: In that case it had more sense since there was a potential variability depending on the minimum fee rate set by the nodes involved in the policy, while here we're changing the model so individual nodes/stakers can't unilaterally change the fee rate so `feeRate` will have probably not many changes.

derekpierre: Right! 👍 . Total it is!

contracts/matic/SubscriptionManager.sol
derekpierre: 👍  - Que barato!

contracts/matic/SubscriptionManager.sol
vzotova: I'm fine with any

vzotova (approved):

scripts/deploy.py
cygnusv: ✅

contracts/matic/SubscriptionManager.sol
cygnusv: ✅

PR #5, opened 2022-02-03
Add support for NPM packages

manumonti:
Adding an NPM package.json file allows the NuCypher contracts to be
treated as an NPM package, which facilitates the import of the contracts
in Node.js frameworks such as Hardhat.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "nucypher-contracts",
+  "version": "0.0.1",
+  "license": "AGPL-3.0-or-later",
+  "description": "NuCypher Network Smart Contracts",
+  "author": "NuCypher"
+}


package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "nucypher-contracts",
+  "version": "0.0.1",
+  "license": "GPL-3.0-or-later"
```
vzotova, line 4: Do we want to have AGPL here?

cygnusv: I think so, our source files specify `// SPDX-License-Identifier: AGPL-3.0-or-later`

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "nucypher-contracts",
+  "version": "0.0.1",
```
cygnusv, line 3: Note that there's an existing "0.1.0" version on https://www.npmjs.com/package/nucypher-contracts, but we may want to just remove it since it would be good to have `nucypher-contracts` on NPM to be organized into the `@nucypher` npm organization.

package.json
manumonti: Threshold-network/solidity-contracts also have this License Identifier.

https://github.com/threshold-network/solidity-contracts/blob/main/package.json

cygnusv, line 4:
```suggestion
  "license": "AGPL-3.0-or-later",
  "description": "NuCypher Network Smart Contracts",
  "author": "NuCypher"
```

cygnusv (changes requested):
Needs proposed changes to the package.json file and rebasing.

package.json
manumonti: ✅ Done!

cygnusv (approved):

package.json
```diff
@@ -0,0 +1,7 @@
...
+  "license": "AGPL-3.0-or-later",
+  "description": "NuCypher Network Smart Contracts",
```
derekpierre, line 5: Side note: we may want to change this to `NuCypher Smart Contracts` or something like that without the `NuCypher Network` combination since we are now Threshold.

PR #21, opened 2022-02-27
Added Solhint and Pylint

### LINKED ISSUE #12: Python and solidity linter GH action

anantasty:
# Linters for Solidity and Python
## Python 
Linting using [Pylint](https://pylint.org/)
Currently no linter configuration is used therefore linter uses default configuration
## Solidity
Uses [Solhint](https://protofire.github.io/solhint/)
Uses default configuration for solidity linting.
We also skip on error because current code with the default configurations does not pass linting.

Fixes #12

### DIFF
--- a/scripts/deploy_subscription_manager.py
+++ b/scripts/deploy_subscription_manager.py
@@ -1,13 +1,15 @@
 #!/usr/bin/python3
 
+from pathlib import Path
+
 import brownie
 from brownie import SubscriptionManager, accounts, project, Contract, Wei
 from brownie.project import compiler
 from brownie.project.main import Project
 
-from pathlib import Path
 
 INITIAL_FEE_RATE = Wei("1 gwei")
+
 
 def main(account_name : str = 'test'):
     deployer = accounts.load(account_name)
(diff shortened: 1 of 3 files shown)


theref:
Hi @anantasty, sorry for the delay in reviewing this. As an initial improvement, could you tidy up the git history a bit? Those 19 commits should probably be squashed down to only one. Thanks!

anantasty:
@theref squashed the commits.

theref (approved):

scripts/deploy_subscription_manager.py
```diff
@@ -30,5 +32,6 @@ def main(account_name : str = 'test'):
         calldata,
         {'from': deployer})
 
-    subscription_manager = Contract.from_abi("SubscriptionManager", transparent_proxy.address, subscription_manager_logic.abi, owner=None)
+    subscription_manager = Contract.from_abi(
+        "SubscriptionManager", transparent_proxy.address, subscription_manager_logic.abi, owner=None)
```
KPrasch, line 36: Controversial opinion:  Linters make some lines more ugly and less readable to follow arbitrary line length rules.  This is an example of that.  Nonetheless this PR does look good to me.  Only suggestion I can make is potentially to just adjust the maximum line length.  whats the average screen resolution in 2022 anyways? 😄

KPrasch (approved):

scripts/deploy_subscription_manager.py
theref: We tend to use black (https://pypi.org/project/black/) everywhere anyway. So if people have `pre-commit` it should catch things before they hit this linter

PR #24, opened 2022-03-29
Use brownie pm fixture to interact with dependencies

### LINKED ISSUE #13: Improve contract resolution to using contracts from dependencies
For the moment, I just found this method to deploy contracts from a dependency:
```python
    brownie_packages = brownie._config._get_data_folder().joinpath('packages')
    oz_dependency_path = 'OpenZeppelin/openzeppelin-contracts@4.5.0/'
    oz = Project("OZ", brownie_packages.joinpath(oz_dependency_path))
 ```
Investigate if there's a better way, in particular, one that doesn't force us to update the path when the dependency version changes

theref:
Fixes #13

Also some cleanup from `black`

### DIFF
--- a/scripts/deploy_subscription_manager.py
+++ b/scripts/deploy_subscription_manager.py
@@ -11,27 +11,19 @@ from brownie.project.main import Project
 INITIAL_FEE_RATE = Wei("1 gwei")
 
 
-def main(account_name : str = 'test'):
+def main(account_name: str = "test"):
     deployer = accounts.load(account_name)
-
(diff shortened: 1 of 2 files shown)


vzotova (approved):

cygnusv:
Thanks for the idea! I didn't know the `pm` fixture existed. There's however a pending aspect here: can we do something similar in a deployment script where we obviously can't use a test fixture?

theref:
I think I've sorted the issue in the deploy script as well

cygnusv (approved):

PR #28, opened 2022-03-30
Replace local import with package import for several libraries

theref:


vzotova (approved):

contracts/contracts/StakingEscrow.sol
```diff
@@ -7,7 +7,7 @@ import "../aragon/interfaces/IERC900History.sol";
 import "./NuCypherToken.sol";
 import "./lib/Bits.sol";
 import "./proxy/Upgradeable.sol";
-import "../zeppelin/math/Math.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
 import "../zeppelin/token/ERC20/SafeERC20.sol";
```
derekpierre, line 11: Any reason this zeppelin contract shouldn't/can't be updated?

theref: I'm working on it, but it causes name collisions when i replace it with a standard import

theref: Fixed

theref:
New update makes some major changes. It removes all references to local OpenZeppelin contracts and moves the override for `approve` into `NuCypherToken.sol`

contracts/contracts/Adjudicator.sol
```diff
@@ -6,8 +6,8 @@ import "./lib/ReEncryptionValidator.sol";
 import "./lib/SignatureVerifier.sol";
 import "./IStakingEscrow.sol";
 import "./proxy/Upgradeable.sol";
-import "../zeppelin/math/SafeMath.sol";
-import "../zeppelin/math/Math.sol";
+import "./lib/AdditionalMath.sol";
```
vzotova, line 9: Probably we don't need AdditionalMath here too

contracts/contracts/lib/AdditionalMath.sol
```diff
@@ -3,14 +3,13 @@
...
 
 
-import "../../zeppelin/math/SafeMath.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
 
 
 /**
 * @notice Additional math operations
 */
 library AdditionalMath {
```
vzotova, line 12: Looks like after removing some contract `AdditionalMath` is also deprecated. Would be great if we can delete too

contracts/contracts/Adjudicator.sol
theref: removed

contracts/contracts/lib/AdditionalMath.sol
theref: removed

vzotova (approved):

PR #59, opened 2022-11-15
DKG Coordinator V1

cygnusv:
Very simple DKG Coordinator contract to start prototyping our DKG protocol. This version of the Coordinator (V1) works as follows:
* We call a DKG protocol execution a `Ritual`.
* States of a `Ritual`:
    * `NON_INITIATED`
    * `WAITING_FOR_TRANSCRIPTS`
    * `WAITING_FOR_CONFIRMATIONS`
    * `ENDED`
* State transitions are based exclusively in time. There's a transcripts window and a confirmations window, defined when the coordinator is deployed.
    * Once a `Ritual` is initiated there's `transcriptsWindow` seconds for the participating nodes to post transcripts
    * Next, there's `confirmationsWindow` seconds to post confirmations
    * After that, the ritual is ended, regardless of the outcome.
* When a `Ritual` is initiated, the caller specifies the nodes that are involved.
    * For the moment, we operate with the fixed assumption that 8 nodes must be specified
* Each DKG `Ritual` gets its own record in the contract, with the following fields:
    * `id`: Just an incremental sequence number
    * `initTimestamp`: timestamp of ritual initiation
    * An array where for each node in the `Ritual` you can find its `Performance`, which is a structure containing transcripts and confirmations data.
* Since transcript data can be large, we don't store it directly in the contract. When "posting transcripts", nodes will pass the transcript bytes as parameters (hence forever stored in the blockchain as TX calldata), and we only store a hash of it as a commitment. This assumes that other nodes and observers are capable of querying the blockchain for the TX data.


contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,123 @@
...
+    event ConfirmationPosted(uint32 indexed ritualID, address indexed node, address[] confirmedNodes);
+
+    enum RitualState {
+        NON_INITIATED,
+        WAITING_FOR_TRANSCRIPTS,
+        WAITING_FOR_CONFIRMATIONS,
+        ENDED
+    }
+
+    // TODO: Find better name
```
theref, line 24: could you explain what a Performance is?

cygnusv: `Performance` would be the information associated to an individual participant of a `Ritual`. In this case, transcript data (technically, just a commitment of this data) and confirmations (i.e., what other nodes in the Ritual have validated them)

contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,123 @@
...
+        bytes32 transcriptDigest = keccak256(transcript);
+        ritual.performance[nodeIndex].transcript = transcriptDigest;
+        
+        emit TranscriptPosted(ritualID, msg.sender, transcriptDigest);
+    }
+
+    function postConfirmation(uint32 ritualID, uint256 nodeIndex, uint256[] calldata confirmedNodesIndexes) external {
+        Ritual storage ritual = rituals[ritualID];
+        require(getRitualState(ritual) == RitualState.WAITING_FOR_CONFIRMATIONS, "Not waiting for confirmations");
+        require(ritual.performance[nodeIndex].transcript != bytes32(0), "Node not part of ritual");
```
theref, line 91: do we need to ensure that `msg.sender` is:
- not the node that is being confirmed
- within the ritual

cygnusv: Right, good catch.

cygnusv: Fixed in 1a7aab7d3f4c1f459188df6c66b9f1a5383165ac. Note that msg.sender can confirm itself explicitly if they want, but the contract already does it when they confirm others, so it's just a waste of gas.

contracts/contracts/Coordinator.sol
piotr-roslaniec: Hmm, maybe something like `Contribution` or `Commitment`?

contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,127 @@
...
+            // TODO: Check nodes are eligible (staking, etc)
+        }
+
+        emit NewRitual(id, nodes);
+    }
+
+    function postTranscript(uint32 ritualID, uint256 nodeIndex, bytes calldata transcript) external {
+        Ritual storage ritual = rituals[ritualID];
+        require(getRitualState(ritual) == RitualState.WAITING_FOR_TRANSCRIPTS, "Not waiting for transcripts");
+        require(ritual.performance[nodeIndex].node == msg.sender, "Node not part of ritual");
```
derekpierre, line 79: Not part of ritual, or incorrect index.

contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,127 @@
...
+    function postConfirmation(uint32 ritualID, uint256 nodeIndex, uint256[] calldata confirmedNodesIndexes) external {
+        Ritual storage ritual = rituals[ritualID];
+        require(getRitualState(ritual) == RitualState.WAITING_FOR_CONFIRMATIONS, "Not waiting for confirmations");
+        require(
+            ritual.performance[nodeIndex].node == msg.sender &&
+            ritual.performance[nodeIndex].transcript != bytes32(0),
+            "Node not part of ritual"
+        );
+
+        require(confirmedNodesIndexes.length <= DKG_SIZE, "Invalid number of confirmations");
```
derekpierre, line 97: Do nodes also confirm themselves, or only (n-1) nodes?

contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,127 @@
...
+* @title CoordinatorV1
+* @notice Coordination layer for DKG-TDec
+*/
+contract CoordinatorV1 {
+
+    uint8 public constant DKG_SIZE = 8;
+
+    event NewRitual(uint32 indexed ritualID, address[] nodes);
+    event TranscriptPosted(uint32 indexed ritualID, address indexed node, bytes32 transcriptDigest);
+    event ConfirmationPosted(uint32 indexed ritualID, address indexed node, address[] confirmedNodes);
```
derekpierre, line 15: Is there any concern about having the list of nodes be publicly known and associated with a specific ritual ID?

contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    function initiateRitual(address[] calldata nodes) external {
+        // TODO: Check for payment
+        // TODO: Check for expiration time
+        // TODO: Improve DKG size choices
+        require(nodes.length == DKG_SIZE, "Invalid number of nodes");
+
+
+        uint32 id = uint32(rituals.length);
+        Ritual storage ritual = rituals.push();
```
derekpierre, line 58: My solidity knowledge is lacking - if this function fails later on (eg. if the nodes aren't sorted properly), is this initialized ritual still stored, or not? If not still stored, great - nothing to see here.

contracts/contracts/Coordinator.sol
cygnusv: They don't have to confirm themselves, that's implicitly assumed if you already posted a transcript, but you could theoretically confirm yourself, there's nothing wrong with that.

contracts/contracts/Coordinator.sol
cygnusv: Yes, but that's still a hard problem, particularly if said nodes will post transcripts and confirmations to the blockchain anyway.

contracts/contracts/Coordinator.sol
cygnusv: Contract calls are atomic in the EVM, so if the call fails due to a revert for example, then there's no state changes.

contracts/contracts/Coordinator.sol
```diff
@@ -0,0 +1,216 @@
...
+    function numberOfRituals() external view returns(uint256) {
+        return rituals.length;
+    }
+
+    function getParticipants(uint32 ritualId) external view returns(Participant[] memory) {
+        Ritual storage ritual = rituals[ritualId];
+        return ritual.participant;
+    }
+
+    function initiateRitual(address[] calldata nodes) external returns (uint32) {
```
theref, line 118: I think this function needs a threshold parameter?

contracts/contracts/Coordinator.sol
KPrasch: I hope we can agree on `Participant` 😄

cygnusv:
Superseded by #75

PR #60, opened 2022-12-12
StakingEscrow: handling special case when staker is SAFT + university

vzotova:


manumonti (approved):
LGTM ✌️

theref (approved):

cygnusv:
For context, last upgrade introduced a minor bug that affected a single staker, whose substake configuration was an edge case we didn't consider. Said upgrade wrongfully locks an amount, which this changeset fixes on an individual basis, for the sake of simplicity and transparency.

derekpierre (commented):
So just to be clear, this change ensures that the staker at the specific address does a cliff vesting (all tokens unlocked) at the original timestamp, as opposed to continuous vesting over time.

vzotova:
> So just to be clear, this change ensures that the staker at the specific address does a cliff vesting (all tokens unlocked) at the original timestamp, as opposed to continuous vesting over time?

These changes add another cliff, second one. First one is when `releaseTimestap == block.timestamp` then all tokens unlocked. before this - only specified value for affected staker

cygnusv:
> So just to be clear, this change ensures that the staker at the specific address does a cliff vesting (all tokens unlocked) at the original timestamp, as opposed to continuous vesting over time?

This staker had a combination of a 2-year linear vesting and a 3-year cliff vesting. The upgrade transformed everything to the latter. Now that the linear vesting is ended, this fix simply reintroduces the original 3-year cliff vesting, so the remaining is assumed to be unlocked.

derekpierre (approved):
🎸

PR #66, opened 2023-02-25
Updates project to use apeworx framework

vzotova:
Brownie -> ape
Main advantage is predictable time travel without random which makes tests more accurate.


scripts/deploy_subscription_manager.py
```diff
@@ -1,24 +1,24 @@
 #!/usr/bin/python3
-from brownie import Contract, SubscriptionManager, Wei, project
-
+from ape import project
 from scripts.utils import get_account
+from web3 import Web3
 
-INITIAL_FEE_RATE = Wei("1 gwei")
+INITIAL_FEE_RATE = Web3.to_wei(1, "gwei")
```
fubuloubu, line 6: Can use `ape.convert(“1 gwei", int)`, or a specific token if you have the `ape_tokens` plugin installed

tests/application/test_operator.py
```diff
@@ -95,12 +93,12 @@ def test_bond_operator(accounts, threshold_staking, pre_application, chain):
...
 
-    pre_application.confirmOperatorAddress({"from": operator1})
+    pre_application.confirmOperatorAddress(sender=operator1)
     assert pre_application.stakingProviderInfo(staking_provider_3)[CONFIRMATION_SLOT]
     assert pre_application.isOperatorConfirmed(operator1)
 
-    assert "OperatorBonded" in tx.events
-    events = tx.events["OperatorBonded"]
+    events = pre_application.OperatorBonded.from_receipt(tx)
+    assert len(events) == 1
```
fubuloubu, line 101:
```suggestion
    assert len(tx.events.filter(pre_application.OperatorBonded)) == 1
```

But look out for 0.6.8 for a better events assertion workflow like this:

```py
assert tx.events == [pre_application.OperatorBonded()]
```

fubuloubu (approved):
Some considerations! Thank you for trying ape!

scripts/deploy_subscription_manager.py
vzotova: I like how it looks, especially with tokens.
I didn't find how to convert data to bytes, int to bytes for example. Could you point where to look at?
And another question: can be used something like that `convert("1_000_000_000 T", int)` in tests before token deployment?

tests/application/test_operator.py
vzotova: `pre_application.OperatorBonded` returns event type. How to make an instance of event? I tried `pre_application.OperatorBonded()` but getting an error that `TypeError: 'ContractEvent' object is not callable`

fubuloubu: The filter approach is the best we can do until the functionality around initializing events is added (hopefully in 0.6.9, sorry I misspoke when I said 0.6.8)

scripts/deploy_subscription_manager.py
fubuloubu: We don't have support for bytes like this right now, it's best to use `eth_utils` for this type of conversion. Our converter is primarily for customized conversions from strings to native python types, prior to abi conversion (per ecosystem)

fubuloubu: Also yes, you should be able to use underscores since we pass it to `Decimal` which should accept that

vzotova: Underscores work, I meant that token not in the list of tokens yet (in test before deploying). In that case how to use `ape-tokens` to run it? (rn I get `ape.exceptions.ConversionError: Must be connected to a provider to use the token converter`) with line `convert("1_000_000_000 NU", int)`

tests/application/test_operator.py
```diff
@@ -14,42 +14,40 @@
...
+import ape
+from web3 import Web3
 from eth_utils import to_checksum_address
 
 CONFIRMATION_SLOT = 1
-MIN_AUTHORIZATION = Wei("40_000 ether")
+MIN_AUTHORIZATION = Web3.to_wei(40_000, "ether")
 MIN_OPERATOR_SECONDS = 24 * 60 * 60
-NULL_ADDRESS = brownie.convert.to_address("0x" + "0" * 40)  # TODO move to some test constants
+NULL_ADDRESS = "0x" + "0" * 40  # TODO move to some test constants
```
banteg, line 24: you can get it from
```python
from ape.utils import ZERO_ADDRESS
```

tests/lib/test_snapshot.py
```diff
@@ -48,25 +48,25 @@ def decode(_snapshot):
...
 
     data = [(block_number + i * 10, value + i) for i in range(10)]
     for i, (block_i, value_i) in enumerate(data):
-        snapshot.addSnapshot(block_i, value_i, {"from": account})
+        snapshot.addSnapshot(block_i, value_i, sender=account)
 
         assert snapshot.length() == i + 1
-        assert snapshot.history(i) == encode(block_i, value_i)
-        assert snapshot.lastSnapshot() == [block_i, value_i]
+        assert snapshot.call_view_method("history", i) == encode(block_i, value_i)
```
banteg, line 63: you can also call non-view methods like this in ape:
```suggestion
        assert snapshot.history.call(i) == encode(block_i, value_i)
```

banteg (approved):

scripts/deploy_subscription_manager.py
```diff
@@ -1,24 +1,24 @@
...
 
 
 def main(id=None):
     deployer = get_account(id)
-    oz = project.load("OpenZeppelin/openzeppelin-contracts@4.5.0/")
+    dependency = project.dependencies["openzeppelin"]["4.8.1"]
+    admin = dependency.get("ProxyAdmin")
+    proxy = dependency.get("TransparentUpgradeableProxy")
```
cygnusv, line 13: @fubuloubu Is there a better way to deploy contracts from a dependency (e.g., like an OZ proxy here?)

ape-config.yaml
```diff
@@ -0,0 +1,54 @@
...
+  #   rinkeby:
+  #     - nu_token: '0x78D591D90a4a768B9D2790deA465D472b6Fe0f18'
+  #       nu_token_supply: 1_000_000_000
+  #       t_staking: '0x18eFb520dA5D387982C860a64855C14C0AcADF3F'
+  #       work_lock: '0x0000000000000000000000000000000000000000'
+  #       staking_escrow: '0x6A6F917a3FF3d33d26BB4743140F205486cD6B4B'
+  #       pre_min_authorization: 40000000000000000000000
+  #       pre_min_operator_seconds: 86400  # one day in seconds
+  #       verify: True
+  #   goerli:
```
theref, line 37: These values are used in the scripts to deploy test networks. Can they be added back in?

tests/application/test_operator.py
vzotova: thanks, just what I wanted!

tests/lib/test_snapshot.py
vzotova: still getting error: `AttributeError: 'AccountHistory' object has no attribute 'call'`

KPrasch (approved):
It's so satisfying to finally see this framework transition... there's a long testing history behind some of the contracts in this repo! nice work 🎉

scripts/deploy_subscription_manager.py
fubuloubu: This should work:

```py
oz = project.dependencies["openzeppelin"]["4.8.1"]
deployer.deploy(oz.ProxyAdmin)
```

tests/lib/test_snapshot.py
fubuloubu: `snapshot._mutable_methods_["history"].call(...)` would be the way to do that when there's a conflict with an ape property

vzotova: yep, we have `history` view method, is current option with `snapshot.call_view_method("history", i)` ok?

fubuloubu: Yes, that works too

scripts/deploy_subscription_manager.py
vzotova: ✔️

tests/application/test_operator.py
vzotova: ✔️

PR #68, opened 2023-03-03
Add config for oryx test network

theref:


derekpierre:
Does this just need a rebase and then a merge?

Issue #69, opened 2023-04-18
L1<->L2 state synchronization


cygnusv:
An implication of our cross-chain coordination contract approach (for reference, 1 stake state contract in Ethereum Mainnet, multiple Coordinator contract in L2s/sidechains), is that we will require certain degree on synchronization between both sides. 

Some examples:
* When initiating a DKG ritual, a L2 Coordinator needs to be sure that the operator correspond to a valid staker (i.e., at least min stake, app is authorized, etc)
* When orderly cohort refreshes occur, e.g., due to a staker unstaking (see https://github.com/nucypher/tdec/issues/123)
* When the result of a ritual produces proof of misbehavior (which can imply slashing on L1)

We need a "common" approach to bridge this communication, taking into consideration also that every L2 may have different messaging strategies.

For example, the L1->L2 step seems to be usually "easy" (i.e., not expensive, and relatively fast): 
* Matic: It only requires to implement some standard methods in the respective Mainnet and Matic contracts, and every call on the mainnet side is bridged to the Matic side when Matic validators sync state. See https://wiki.polygon.technology/docs/develop/l1-l2-communication/state-transfer#state-transfer-from-ethereum--polygon
* Optimism: Similar approach, syncing takes approximately 15 minutes. See https://community.optimism.io/docs/developers/bridge/messaging/#for-ethereum-l1-to-optimism-l2-transactions

The L2->L1 step it's also possible but seems more involved, and I'd say we want to find approaches that reduce this communication as possible.

cygnusv:
Tentative design, with the goal of abstracting away cross-chain complexity from both the CBD application contract and the L2 Coordinators. Complexity is hidden the intermediary `StateSyncRoot` and `StateSyncChild` contracts. See this (awful) figure, if it helps:
<img width="1344" alt="image" src="https://user-images.githubusercontent.com/2564234/232802749-c8c30927-2296-46f2-8509-f81155b48876.png">

This design would imply the deployment of a pair of contracts for each layer 2 (e.g., for Matic:`StateSyncRootMatic` on Mainnet and `StateSyncChildMatic` on Matic), where each pair implements the most suitable cross-chain mechanism for that particular layer 2.

The CBD application contract would have a mapping of `uint256 => StateSyncRoot`, with chainIDs as keys.
Each L2 Coordinator contract would have a static reference to their corresponding `StateSyncChild`.

cygnusv:
[...]

PR #75, opened 2023-04-26
Prepare Coordinator contract to get external stakes info

cygnusv:
* This is in preparation of stake information synchronization from Ethereum to L2s (see @theref's incoming work here https://github.com/cygnusv/nucypher-contracts/pull/6)
* Introduces a simple BLS12381 library contract to deal with serialized G1 and G2 points
* Also ports Coordinator tests to ape


contracts/contracts/coordination/Coordinator.sol
```diff
@@ -41,40 +36,35 @@ contract Coordinator is Ownable {
...
         uint32 dkgSize;
         uint32 initTimestamp;
         uint32 totalTranscripts;
         uint32 totalAggregations;
-        bytes32 aggregatedTranscriptHash;
+        BLS12381.G1Point publicKey;
         bool aggregationMismatch;
         bytes aggregatedTranscript;
-        bytes publicKey;
-        bytes32 publicKeyHash;
```
derekpierre, line 64: Thanks for getting rid of my bloated values 😅

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -18,11 +13,11 @@ interface ApplicationInterface {
 contract Coordinator is Ownable {
 
     // Ritual
-    event StartRitual(uint32 indexed ritualId, address indexed initiator, address[] nodes);
+    event StartRitual(uint32 indexed ritualId, address indexed initiator, address[] participants);
     event StartTranscriptRound(uint32 indexed ritualId);
```
derekpierre, line 17: Not necessarily related to this PR, but @KPrasch and I were thinking that these two events could be combined into one. It's just superfluous to have the two of them. 

```
        emit StartRitual(id, msg.sender, providers);
        emit StartTranscriptRound(id);
```

cygnusv: It's a possibility, but I see those events to be different because they have a different audience: `StartTranscriptRound` is almost exclusively for nodes (and perhaps monitoring tools), and `StartRitual` is to be consumed by adopters, nodes and external observers.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -125,63 +115,55 @@ contract Coordinator is Ownable {
...
-            previousNode = currentNode;
-            // TODO: Check nodes are eligible (staking, etc)
+            newParticipant.provider = current;
+            previous = current;
         }
-        // TODO: Compute cohort fingerprint as hash(nodes)
-
-        emit StartRitual(id, msg.sender, nodes);
+        
+        // TODO: Include cohort fingerprint in StartRitual event?
```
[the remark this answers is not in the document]
cygnusv: Cohort fingerprints can be used as a cohort unique identifier. It may be useful (or not 😅 )

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -41,40 +36,35 @@ contract Coordinator is Ownable {
...
     // TODO: Optimize layout
     struct Ritual {
-        uint32 id;  // TODO: Redundant? ID is index of rituals array
         address initiator;
         uint32 dkgSize;
         uint32 initTimestamp;
         uint32 totalTranscripts;
         uint32 totalAggregations;
-        bytes32 aggregatedTranscriptHash;
+        BLS12381.G1Point publicKey;
```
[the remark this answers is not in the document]
cygnusv: correct, that's how it should be, but we need to double-check there are no other serialization artifacts (e.g., array lengths). For reference, this length should be exactly 48 bytes.

contracts/contracts/coordination/Coordinator.sol
derekpierre: Sure, but there is overlap, and I'm not sure how much value just `ritual id` by itself provides.

`StartRitual` is the richer one since the `providers` value is included in it. This is more valuable than `StartTranscriptRound`, because nodes can easily check whether they are part of the ritual's cohort or not directly from the event. Getting an additional `StartTranscriptRound` in addition to `StartRitual` event doesn't add any real value.

Initially, when we mused about a check-in period between ritual initialization but before posting transcripts, it was valuable,  but I'm not so convinced anymore since we don't do any check-ins.

derekpierre (approved):
Great PR! 🎸

vzotova (approved):
LGTM
One question: `StakeInfo.sol` is it part of test contracts?

cygnusv:
> LGTM One question: `StakeInfo.sol` is it part of test contracts?

No, it's in fact an important piece of the synchronization between the CBD app contract in L1 and the coordinators in L2s (https://github.com/nucypher/nucypher-contracts/issues/69)

But it's also great for tests and testnets, because you only need to grant the update role to be able to set stake information

PR #76, opened 2023-05-03
Add cross chain messaging for stake info from ETH to Polygon

theref:


theref:
replaced by https://github.com/nucypher/nucypher-contracts/pull/76

contracts/contracts/coordination/StakeInfo.sol
```diff
@@ -13,17 +13,19 @@ import "../../threshold/IAccessControlApplication.sol";
...
         uint96 amount;
         // TODO: what about undelegations etc?
     }
 
-    constructor(address[] memory updaters){
+    constructor(address _polygonChild, address[] memory updaters){
         for(uint i = 0; i < updaters.length; i++){
             _grantRole(UPDATE_ROLE, updaters[i]);
         }
+        polygonChild = _polygonChild;
```
cygnusv, line 28: The polygon child is an updater, so there's no need to track it. An access-control based design allows the StakeInfo contract to be completely unaware of how the update mechanism works: it could be a cross-chain messaging contract, but also could be a multisig or a privileged EOA.

contracts/xchain/PolygonChild.sol
```diff
@@ -0,0 +1,27 @@
...
+        (bool success, /* returnId */ ) = stakeInfoAddress.call(data);
+    }
+
+    function sendMessageToRoot(bytes memory message) public {
+        _sendMessageToRoot(message);
+    }
+
+    function setStakeInfoAddress(address _stakeInfoAddress) public onlyOwner {
+        stakeInfoAddress = _stakeInfoAddress;
+    }
```
cygnusv, line 26: Can we move this to a constructor and make the contract not Ownable?

theref: then we have circular dependencies. PolygonChild will take StakeInfo address in the constructor, and StakeInfo will take PolygonChild address in the constructor (in the updators)

cygnusv: Ah good point. Let's keep StakeInfo independent then 👍

contracts/xchain/PolygonChild.sol
```diff
@@ -0,0 +1,27 @@
...
+    address public stakeInfoAddress;
+    
+    constructor(address _fxChild) FxBaseChildTunnel(_fxChild) {}
+
+    function _processMessageFromRoot(
+        uint256 /* stateId */,
+        address sender,
+        bytes memory data
+    ) internal override validateSender(sender) {
+        (bool success, /* returnId */ ) = stakeInfoAddress.call(data);
```
cygnusv, line 17: Should we have some sort of event?

KPrasch:
Looks like this PR needs a minor rebase!

scripts/check_xchain.py
```diff
@@ -0,0 +1,20 @@
...
+        root = project.PolygonRoot.at("0xdc90A337DF9561705EB85B92391ab8F55d114D53")
+        root.updateOperator(
+            "0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600",
+            "0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600",
+        )
+    time.sleep(60 * 40)
+    with networks.polygon.mumbai.use_provider("infura"):
+        stake_info = project.StakeInfo.at("0x40D0107ACa3503CB345E4117a9F92E8220EEEb3C")
+        print(stake_info.operatorToProvider("0xAe87D865F3A507185656aD0ef52a8E0B9f3d58f8"))
+        print(stake_info.operatorToProvider("0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600"))
```
KPrasch, line 20: This appears to be an (excellent) smoke test.  Is this a sufficient degree of testing or is there something we can do in local automated tests?

theref: i haven't come across anyone doing automated tests properly. @cygnusv and I came the conclusion that the benefit gained from having the tests probably doesn't out-weigh the time spent getting them to work

KPrasch (approved):

PR #77, opened 2023-05-04
Add a public key field to the Participant struct

piotr-roslaniec:
- Ferveo's public keys are required to perform aggregate verification
- Added a mapping from the provider address to `FerveoPublicKey` bytes
- Downstream changes required: 
  - https://github.com/nucypher/nucypher-ts/issues/234
  - https://github.com/nucypher/nucypher/issues/3167
- Documented possible extension in: https://github.com/nucypher/nucypher-contracts/issues/95
- Required reviews: 2

### DIFF
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -30,6 +30,8 @@ contract Coordinator is AccessControlDefaultAdminRules {
     event TimeoutChanged(uint32 oldTimeout, uint32 newTimeout);
     event MaxDkgSizeChanged(uint16 oldSize, uint16 newSize);
 
+    event ParticipantPublicKeySet(address indexed participant, BLS12381.G1Point publicKey);
+
     enum RitualState {
         NON_INITIATED,
         AWAITING_TRANSCRIPTS,
@@ -61,11 +63,16 @@ contract Coordinator is AccessControlDefaultAdminRules {
         Participant[] participant;
     }
 
+    struct ParticipantKey {
+        uint32 ritualId;
+        BLS12381.G1Point publicKey;
+    }
+
     using SafeERC20 for IERC20;
 
     bytes32 public constant INITIATOR_ROLE = keccak256("INITIATOR_ROLE");
 
-    mapping(address => bytes) public providerPublicKey;
+    mapping(address => ParticipantKey[]) keysHistory;
 
     IAccessControlApplication public immutable application;
 
@@ -101,13 +108,13 @@ contract Coordinator is AccessControlDefaultAdminRules {
     function getRitualState(Ritual storage ritual) internal view returns (RitualState){
         uint32 t0 = ritual.initTimestamp;
         uint32 deadline = t0 + timeout;
-        if (t0 == 0){
+        if (t0 == 0) {
             return RitualState.NON_INITIATED;
         } else if (ritual.totalAggregations == ritual.dkgSize) {
             return RitualState.FINALIZED;
-        } else if (ritual.aggregationMismatch){
+        } else if (ritual.aggregationMismatch) {
             return RitualState.INVALID;
-        } else if (block.timestamp > deadline){
+        } else if (block.timestamp > deadline) {
             return RitualState.TIMEOUT;
         } else if (ritual.totalTranscripts < ritual.dkgSize) {
             return RitualState.AWAITING_TRANSCRIPTS;
@@ -126,10 +133,26 @@ contract Coordinator is AccessControlDefaultAdminRules {
         _setRoleAdmin(INITIATOR_ROLE, bytes32(0));
     }
 
-    function setProviderPublicKey(bytes calldata publicKey) external {
-        // TODO: Verify public key length
-        require(publicKey.length == 48, "Invalid public key length");
-        providerPublicKey[msg.sender] = publicKey;
+    function setProviderPublicKey(BLS12381.G1Point calldata _publicKey) public {
+        uint32 lastRitualId = uint32(rituals.length);
+        address provider = application.stakingProviderFromOperator(msg.sender);
+
+        ParticipantKey memory newRecord = ParticipantKey(lastRitualId, _publicKey);
+        keysHistory[provider].push(newRecord);
+
+        emit ParticipantPublicKeySet(provider, _publicKey);
+    }
+
+    function getProviderPublicKey(address _address, uint _ritualId) public view returns (BLS12381.G1Point memory) {
+        ParticipantKey[] storage participantHistory = keysHistory[_address];
+
+        for (uint i = participantHistory.length - 1; i >= 0; i--) {
+            if (participantHistory[i].ritualId <= _ritualId) {
+                return participantHistory[i].publicKey;
+            }
+        }
+
+        revert("No keys found prior to the provided ritual");
     }
 
     function setTimeout(uint32 newTimeout) external onlyRole(DEFAULT_ADMIN_ROLE) {
@@ -152,11 +175,11 @@ contract Coordinator is AccessControlDefaultAdminRules {
         // TODO: Events
     }
 
-    function numberOfRituals() external view returns(uint256) {
+    function numberOfRituals() external view returns (uint256) {
         return rituals.length;
     }
 
-    function getParticipants(uint32 ritualId) external view returns(Participant[] memory) {
+    function getParticipants(uint32 ritualId) external view returns (Participant[] memory) {
         Ritual storage ritual = rituals[ritualId];
         return ritual.participant;
     }
@@ -184,19 +207,18 @@ contract Coordinator is AccessControlDefaultAdminRules {
         ritual.endTimestamp = ritual.initTimestamp + duration;
 
         address previous = address(0);
-        for(uint256 i=0; i < length; i++){
+        for (uint256 i = 0; i < length; i++) {
             Participant storage newParticipant = ritual.participant.push();
             address current = providers[i];
             // Make sure that current provider has already set their public key
-            require(
-                providerPublicKey[current].length > 0,
-                "Provider has not set their public key"
-            );
+            ParticipantKey[] storage participantHistory = keysHistory[current];
+            require(participantHistory.length > 0, "Provider has not set their public key");
+
             require(previous < current, "Providers must be sorted");
             // TODO: Improve check for eligible nodes (staking, etc) - nucypher#3109
             // TODO: Change check to isAuthorized(), without amount
             require(
-                application.authorizedStake(current) > 0, 
+                application.authorizedStake(current) > 0,
                 "Not enough authorization"
             );
             newParticipant.provider = current;
@@ -210,7 +232,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
         return id;
     }
 
-    function cohortFingerprint(address[] calldata nodes) public pure returns(bytes32) {
+    function cohortFingerprint(address[] calldata nodes) public pure returns (bytes32) {
         return keccak256(abi.encode(nodes));
     }
 
@@ -244,7 +266,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
         ritual.totalTranscripts++;
 
         // end round
-        if (ritual.totalTranscripts == ritual.dkgSize){
+        if (ritual.totalTranscripts == ritual.dkgSize) {
             emit StartAggregationRound(ritualId);
         }
         processReimbursement(initialGasLeft);
@@ -296,9 +318,9 @@ contract Coordinator is AccessControlDefaultAdminRules {
             ritual.aggregatedTranscript = aggregatedTranscript;
             ritual.publicKey = publicKey;
         } else if (
-            !BLS12381.eqG1Point(ritual.publicKey, publicKey) || 
-            keccak256(ritual.aggregatedTranscript) != aggregatedTranscriptDigest
-        ){
+            !BLS12381.eqG1Point(ritual.publicKey, publicKey) ||
+        keccak256(ritual.aggregatedTranscript) != aggregatedTranscriptDigest
+        ) {
             ritual.aggregationMismatch = true;
             emit EndRitual({
                 ritualId: ritualId,
@@ -306,9 +328,9 @@ contract Coordinator is AccessControlDefaultAdminRules {
             });
         }
 
-        if(!ritual.aggregationMismatch){
+        if (!ritual.aggregationMismatch) {
             ritual.totalAggregations++;
-            if (ritual.totalAggregations == ritual.dkgSize){
+            if (ritual.totalAggregations == ritual.dkgSize) {
                 processPendingFee(ritualId);
                 emit EndRitual({
                     ritualId: ritualId,
@@ -327,9 +349,9 @@ contract Coordinator is AccessControlDefaultAdminRules {
     ) internal view returns (Participant storage) {
         uint length = ritual.participant.length;
         // TODO: Improve with binary search
-        for(uint i = 0; i < length; i++){
+        for (uint i = 0; i < length; i++) {
             Participant storage participant = ritual.participant[i];
-            if(participant.provider == provider){
+            if (participant.provider == provider) {
                 return participant;
             }
         }
@@ -345,7 +367,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
 
     function processRitualPayment(uint256 ritualID, address[] calldata providers, uint32 duration) internal {
         uint256 ritualCost = feeModel.getRitualInitiationCost(providers, duration);
-        if (ritualCost > 0){
+        if (ritualCost > 0) {
             totalPendingFees += ritualCost;
             assert(pendingFees[ritualID] == 0);  // TODO: This is an invariant, not sure if actually needed
             pendingFees[ritualID] += ritualCost;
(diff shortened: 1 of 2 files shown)


contracts/contracts/coordination/Coordinator.sol
```diff
@@ -39,6 +39,7 @@ contract Coordinator is Ownable {
         address provider;
         bool aggregated;
         bytes transcript;  // TODO: Consider event processing complexity vs storage cost
+        bytes publicKey;  // TODO: Consider event processing complexity vs storage cost
```
derekpierre, line 42: I know this PR is still WIP, but looking forward to <https://github.com/nucypher/nucypher/issues/3081>, can we call this `blinding key` or something more specific? There will need to be another entry in here for the E2E protocol for something that equates to a `decryption request encrypting key` - see https://github.com/nucypher/nucypher/issues/3081#issuecomment-1523814876.

KPrasch:
I'm curious what's compelling us to put ferveo public keys onchain specifically?  As of `nucypher@dkg-dev-2` these keys are known to participants by using P2P/node discovery which can be interpreted as a duplicated effort?

piotr-roslaniec:
> are known to participants by using P2P/node discovery

We also need to make this information (`[Transcript, PublicKey]` pair) available to anyone who wants to verify the aggregate (Bob).

derekpierre:
> We also need to make this information ([Transcript, PublicKey] pairs) available to anyone who wants to verify the aggregate (Bob).

The ability to verify without having to learn about the network is advantageous. Learning is a very involved process. A web-based Bob would be able to verify without needing to learn. 

@KPrasch it would also save us (on the Python-side) from needing Ursula objects for this code: https://github.com/nucypher/nucypher/blob/development/nucypher/characters/lawful.py#L696

piotr-roslaniec:
Notes from a discussion on Discord voice:
- It's not possible to counterfeit `(Transcript, PublicKey)` pair. And so, we only need to provide DA for the `Transcript`. `PublicKey` can provide separately without compromising the security of aggregate verification.
- Putting `PublicKey` on-chain is a naive design
- An alternative design: Make it available through Porter, and when Porter is deprecated, use the same method od `PublicKey` discovery to communicate directly with nodes

@KPrasch @derekpierre

KPrasch:
Another consideration here is if we want to derive new keys for each participant+ritual.  Otherwise, if nodes only have one long-term ferveo keypair, there's no sense it writing it to the Participant struct.

piotr-roslaniec:
>  Otherwise, if nodes only have one long-term ferveo keypair, there's no sense it writing it to the Participant struct.

In that case, mapping could be better (updated).

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -127,8 +127,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
     }
 
     function setProviderPublicKey(bytes calldata publicKey) external {
-        // TODO: Verify public key length
-        require(publicKey.length == 48, "Invalid public key length");
+        require(publicKey.length == 104, "Invalid public key length");
```
cygnusv, line 130: We have G2 element size in here:
https://github.com/nucypher/nucypher-contracts/blob/da35b9d9f13ddebd7ce7bbbc3fbc7d5fb0d95411/contracts/contracts/lib/BLS12381.sol#L11
I guess this 104 comes from the usual extra 8 bytes for length from Rust. What do you think about using the `G2Point` struct from our `BLS12381` lib?

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -124,6 +126,12 @@ contract Coordinator is AccessControlDefaultAdminRules {
         _setRoleAdmin(INITIATOR_ROLE, bytes32(0));
     }
 
+    function setProviderPublicKey(bytes calldata publicKey) external {
+        // TODO: Verify public key length
+        require(publicKey.length == 48, "Invalid public key length");
+        providerPublicKey[msg.sender] = publicKey;
```
cygnusv, line 131: This should be a mapping from providers to keys, not from operators. You can get the provider address this way:
```js
address provider = application.stakingProviderFromOperator(msg.sender);
```

contracts/contracts/coordination/Coordinator.sol
piotr-roslaniec: Sounds good, I should update serialization in `ferveo-py` bindings then

piotr-roslaniec: Ferveo changes: https://github.com/nucypher/ferveo/pull/136

contracts/contracts/coordination/Coordinator.sol
piotr-roslaniec: :heavy_check_mark:

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -124,6 +126,11 @@ contract Coordinator is AccessControlDefaultAdminRules {
         _setRoleAdmin(INITIATOR_ROLE, bytes32(0));
     }
 
+    function setProviderPublicKey(BLS12381.G2Point calldata publicKey) external {
```
theref, line 129: does this need some access control?

piotr-roslaniec: Good question. One can only set a public key for themselves so I assumed it's one of the cases when we don't necessarily need AC

cygnusv: Agreed

piotr-roslaniec:
I just need 2 approvals so I'm going to ahead and merge if that sounds ok

contracts/contracts/coordination/Coordinator.sol
derekpierre: Would we ever consider the case that Ursulas would have a different key per ritual? Is there any advantage to that?

Does this model not allow the key to change at all...

cygnusv: > Would we ever consider the case that Ursulas would have a different key per ritual? Is there any advantage to that?

I think so, but IIRC, we currently only have long-term Ferveo keys, so for the moment this works.

> Does this model not allow the key to change at all... eg. if it was compromised etc.

Isn't the opposite? This model allow to set the key at any time.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -65,6 +65,8 @@ contract Coordinator is AccessControlDefaultAdminRules {
 
     bytes32 public constant INITIATOR_ROLE = keccak256("INITIATOR_ROLE");
 
+    mapping(address => BLS12381.G2Point) public providerPublicKey;
```
vzotova, line 68: I'm out of context, why here provider used instead of operator?

cygnusv: Because the mapping key should be the provider not the operator (although technically the key is generated by the node)

contracts/contracts/coordination/Coordinator.sol
cygnusv: OK I see what you mean. Well, the very basic requirement here is to have a minimal form of commitment and retrievability of the Ferveo key, since for the moment we are not going to be able to do any on-chain verification. This means that, even if keys are changed at a given time, there's always a record of the previous keys (not accessible via contract state, but accessible through the TX hash or historical data). For our purposes, I think it could suffice. 

Now, if we want to improve this in line with your comment, I think it wouldn't take much effort. We could do something like, e.g., an append-only list where for each node the Coordinator stores a pair of `(public_key, start_ritual)`, so in the optimistic case this list is of length 1.

contracts/contracts/coordination/Coordinator.sol
vzotova: That's my question why provider key? Provider is technically staker and I wonder why staker needed

cygnusv: We keep the model where the node identifier from a protocol point of view is the staker (staking provider if we're exact), similar to how NuCypher works.

contracts/contracts/coordination/Coordinator.sol
piotr-roslaniec: I think for the purpose of local verification emitting `ParticipantPublicKeySet(public_key)` event would also be acceptable. Local verification should be performed at least once per ritual and searching through events shouldn't be too much of a hassle in this case. 

For on-chain verification mapping is sufficient and no access to historic public key values is needed.

derekpierre: > This means that, even if keys are changed at a given time, there's always a record of the previous keys (not accessible via contract state, but accessible through the TX hash or historical data).

Right.

Nice thought! If we accept that sleuthing blockchain information is acceptable for verification an event could be a good solution 👍

contracts/contracts/coordination/Coordinator.sol
derekpierre: Perhaps this should be plural `providerPublicKeys` instead of `providerPublicKey`?

I guess presumably the expectation is that on start-up, Ursulas would need to check that they have already submitted the relevant public key or update it if not? This would be before starting to process any ritual events. (cc @KPrasch )

piotr-roslaniec: Documented downstream changs in `nucypher` here: https://github.com/nucypher/nucypher/issues/3167

cygnusv: > Perhaps this should be plural `providerPublicKeys` instead of `providerPublicKey`?

I don't know, I personally tend to use singular for arrays and mappings. But I'm fine with adopting a common convention for this.

contracts/contracts/coordination/Coordinator.sol
KPrasch: > Would we ever consider the case that Ursulas would have a different key per ritual? Is there any advantage to that?
> 
> Does this model not allow the key to change at all... eg. if it was compromised etc.

This is also my impression, there's no harm in supporting ferveo public keys in the Ritual struct despite ursula's only using a long-term public key at present. The downstream logic is easier to understand and implement imo.

derekpierre: > ParticipantPublicKeySet(public_key) event would also be acceptable

Rethinking this I'm not sure that this is really that feasible. For a Cohort say of size 16-of-30, how feasible is it to possibly look for / process these events. I'm thinking with respect to @piotr-roslaniec 's local verification PR, that would need a lot more logic to accommodate these events.

> This is also my impression, there's no harm in supporting ferveo public keys in the Ritual struct

This was my first inclination as well but I can see the argument that it involves extra contract storage space and additional gas costs which may be mostly wasted *if* the key never changes.

> e.g., an append-only list where for each node the Coordinator stores a pair of (public_key, start_ritual), so in the optimistic case this list is of length 1.

Maybe we're back to @cygnusv 's solution and perhaps if we use this type of storage, then we provide a getter function in the contract that takes a ritual number and a provider address, and returns the relevant key...maybe...? (cc @vzotova )

piotr-roslaniec: > For a Cohort say of size 16-of-30, how feasible is it to possibly look for / process these events

I don't think it's too bad: Knowing the ritual timestamp, for each participant we need to find the "last-before" event. I don't see this as an issue because I assumed that participants wouldn't change their keys very often and that the filtering is done over JSON-RPC anyway.

`(public_key, start_ritual)` sounds good to me.

piotr-roslaniec: @cygnusv Does this implementation sketch make sense?

```
    mapping(address => KeyRecord[]) keysHistory;
    event ParticipantPublicKeySet(address indexed participant, BLS12381.G2Point publicKey);
    
    function setPublicKey(bytes calldata _publicKey) public {
        BLS12381.G2Point memory publicKey = BLS12381.bytesToG2Point(_publicKey);
        uint startRitual = block.timestamp;
        
        KeyRecord memory newRecord = KeyRecord(publicKey, startRitual);
        keysHistory[msg.sender].push(newRecord);
        
        emit ParticipantPublicKeySet(msg.sender, publicKey);
    }

    function getPreviousKey(address _address, uint _ritual) public view returns (BLS12381.G2Point memory) {
        KeyRecord[] storage userHistory = keysHistory[_address];
        require(userHistory.length > 0, 'No keys found for the provided address');

        for(uint i = userHistory.length - 1; i >= 0; i--) {
            if(userHistory[i].startRitual <= _ritual) return userHistory[i].publicKey;
        }
        
        revert('No keys found prior to the provided ritual');
    }
    ```

Comment (cygnusv) [reply contracts/contracts/coordination/Coordinator.sol]: Looks good in genera, but some comments: 
* Why `uint startRitual = block.timestamp;`? Can we just do `rituals.length` which is the ID of the next ritual?
* `msg.sender` is the operator, so we need to resolve the provider. 
* We could remove the `require(participantHistory.length > 0, 'No keys found for the provided address');`

File: contracts/contracts/coordination/Coordinator.sol [hunk 6]
```diff
@@ -61,10 +63,17 @@ contract Coordinator is AccessControlDefaultAdminRules {
         Participant[] participant;
     }
 
+    struct ParticipantKey {
+        uint32 ritualId;
```
derekpierre, line 67: Is this more like `fromRitualId` or ... something similar?

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -124,6 +133,28 @@ contract Coordinator is AccessControlDefaultAdminRules {
...
+        uint32 lastRitualId = uint32(rituals.length);
+        address provider = application.stakingProviderFromOperator(msg.sender);
+
+        ParticipantKey memory newRecord = ParticipantKey(lastRitualId, _publicKey);
+        keysHistory[provider].push(newRecord);
+
+        emit ParticipantPublicKeySet(provider, _publicKey);
+    }
+
+    function getProviderPublicKey(address _address, uint _ritualId) public view returns (BLS12381.G1Point memory) {
```
derekpierre, line 146: If this function is not called by the other functions in the contract i.e. only called from an external entitiy, I believe using `external` instead saves a copy step of arguments.

Also, let's call the parameter `_provider`.

```suggestion
    function getProviderPublicKey(address _provider, uint _ritualId) external view returns (BLS12381.G1Point memory) {
```

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -30,6 +30,8 @@ contract Coordinator is AccessControlDefaultAdminRules {
     event TimeoutChanged(uint32 oldTimeout, uint32 newTimeout);
     event MaxDkgSizeChanged(uint16 oldSize, uint16 newSize);
 
+    event ParticipantPublicKeySet(address indexed participant, BLS12381.G1Point publicKey);
```
cygnusv, line 33: Isn't this a G2 point?

tests/test_coordinator.py
```diff
@@ -130,10 +143,22 @@ def test_initiate_ritual(coordinator, nodes, initiator, erc20, flat_rate_fee_mod
     assert coordinator.getRitualState(0) == RitualState.AWAITING_TRANSCRIPTS
 
 
+def test_test_provider_public_key(coordinator, nodes):
```
derekpierre, line 146: Typo?:

```suggestion
def test_provider_public_key(coordinator, nodes):
```

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -124,6 +133,28 @@ contract Coordinator is AccessControlDefaultAdminRules {
...
     }
 
+    function setProviderPublicKey(BLS12381.G1Point calldata _publicKey) public {
+        uint32 lastRitualId = uint32(rituals.length);
+        address provider = application.stakingProviderFromOperator(msg.sender);
+
+        ParticipantKey memory newRecord = ParticipantKey(lastRitualId, _publicKey);
+        keysHistory[provider].push(newRecord);
+
+        emit ParticipantPublicKeySet(provider, _publicKey);
```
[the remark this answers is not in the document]
piotr-roslaniec: Could simplify filtering
:heavy_check_mark:

tests/test_coordinator.py
piotr-roslaniec: :heavy_check_mark:

contracts/contracts/coordination/Coordinator.sol
piotr-roslaniec: :heavy_check_mark:

KPrasch (approved):
👨🏻‍🚀

cygnusv (approved):
Great work @piotr-roslaniec. An additional idea from a discussion with @KPrasch: what a about adding the ritual public key as part of the post transcript TX? 

* Piggy-backing from the post transcript TX simplifies things, maybe we can make it an optional parameter in case the the node needs to set the key. Less TXs means simpler client code, less problems.
* Prepares API for when/if Ferveo keys are generated per-ritual.
* Code definitely would look uglier, but I think this is bearable considering Coordinator is for the private beta release.

If this sounds good, we can do this in a separate PR, no need to do it here.

derekpierre (approved):
🎸

piotr-roslaniec:
Documented possible extension in: https://github.com/nucypher/nucypher-contracts/issues/95

jdbertron:
> Another consideration here is if we want to derive new keys for each participant+ritual. Otherwise, if nodes only have one long-term ferveo keypair, there's no sense it writing it to the Participant struct.

I don't follow.

PR #83, opened 2023-05-24
Updates coordinator deployment scripting

KPrasch:


theref (approved):

scripts/verify_contract.py
```diff
@@ -0,0 +1,4 @@
+from ape import networks
+
+etherscan = networks.provider.network.explorer
+etherscan.publish_contract("0xc7590Ba497d1A7AB82c1BD61f7Dcd2685990986e")
```
derekpierre, line 4: Should the contract address be configurable?

KPrasch: Absolutely.  I couldn't get this working on polygonscan so I stopped short here.

theref: Yeah there is a bug at the mo. But they have been able to reproduce it and a fix is on the way https://discord.com/channels/922917176040640612/954593641970696292/1109206518286528613

derekpierre (approved):
🎸

PR #86, opened 2023-06-09
Beefing up the Coordinator: Restrictions to initiator, ritual duration, fee model, automatic TX reimbursements, etc

cygnusv:
This is an omnibus PR that addresses several pending functionalities for the Coordinator. From a subjective order of relevance:
1. Introduces **fee models**: initially we're just introducing here a basic `IFeeModel` interface that allows to calculate initiation costs, but that could be later extended to manage other payment inputs. For this PR, I'm instantiating the interface with a simple `FlatRateFeeModel`, where ritual initiation cost is just `dkg_size *  duration * rate` and the currency is assumed to be an ERC20 token. We also introduce `duration` as a parameter to initiate rituals.
2. Introduces an **access-controlled role for initiating rituals**, as well as a one-off switch to unleash ritual initiation. The proposed model works as follows:
    - Initially, Coordinator contract works in permissioned way (only addresses that have the `INITIATOR_ROLE` can create rituals)
    - This role can be granted by the Coordinator admin (uses OpenZeppelin access control model, see tests)
    - Admin can switch the Coordinator from permissioned to permissionless ritual initiation; at this point, the initiator role is not relevant anymore. Note that this action can't be undone.
    - Rationale for this model is that it both works for many types of deployments without changing the code:
         - It's the exact model we plan for the beta release (initially, ritual initiation is controlled by Threshold DAO but can be opened in the future)
         - For devnets and testnets, we have maximum flexibility: if we want it permissioned, we just grant initiator roles; if we want permissionless, we just call the switch.
3. Automatic **reimbursement of costs** for posting transcripts and aggregations. It uses Keep's ReimbursementPool model, currently in production with tBTCv2
4. Ensures that provided decryption request static key by nodes is 42 byte (courtesy of @derekpierre  🎸 )

### DIFF
--- a/ape-config.yaml
+++ b/ape-config.yaml
@@ -11,10 +11,10 @@ plugins:
 dependencies:
   - name: openzeppelin
     github: OpenZeppelin/openzeppelin-contracts
-    version: 4.8.1
+    version: 4.9.1
   - name: openzeppelin-upgradeable
     github: OpenZeppelin/openzeppelin-contracts-upgradeable
-    version: 4.8.1
+    version: 4.9.1
   - name: fx-portal
     github: 0xPolygon/fx-portal
     version: 1.0.5
@@ -23,8 +23,8 @@ solidity:
   version: 0.8.20
   evm_version: paris
   import_remapping:
-    - "@openzeppelin/contracts=openzeppelin/v4.8.1"
-    - "@openzeppelin-upgradeable/contracts=openzeppelin-upgradeable/v4.8.1"
+    - "@openzeppelin/contracts=openzeppelin/v4.9.1"
+    - "@openzeppelin-upgradeable/contracts=openzeppelin-upgradeable/v4.9.1"
     - "@fx-portal/contracts=fx-portal/v1.0.5"
 
 deployments:
--- a/ape-config.yaml
+++ b/ape-config.yaml
@@ -29,7 +29,14 @@ solidity:
 
 deployments:
   polygon:
+    mainnet:
+      - contract_type: DAI
+        address: '0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063'
     mumbai:
+      - contract_type: DAI
+        address: '0x001B3B4d0F3714Ca98ba10F6042DaEbF0B1B7b6F'
+      - contract_type: StakeInfo
+        address: '0xC1379866Fb0c100DCBFAb7b470009C4827D47DD8'
       - fx_child: '0xCf73231F28B7331BBe3124B907840A94851f9f11'
       - verify: False
   ethereum:
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -262,13 +262,18 @@ contract Coordinator is AccessControlDefaultAdminRules {
 
         require(
             participant.decryptionRequestStaticKey.length == 0,
-            "Node already provided request encrypting key"
+            "Node already provided decryption request static key"
+        );
+
+        require(
+            decryptionRequestStaticKey.length == 42,
+            "Invalid length for decryption request static key"
         );
 
         // nodes commit to their aggregation result
         bytes32 aggregatedTranscriptDigest = keccak256(aggregatedTranscript);
         participant.aggregated = true;
-        participant.decryptionRequestStaticKey = decryptionRequestStaticKey;  // TODO validation?
+        participant.decryptionRequestStaticKey = decryptionRequestStaticKey;
         emit AggregationPosted(ritualId, provider, aggregatedTranscriptDigest);
 
         if (ritual.aggregatedTranscript.length == 0) {
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -5,6 +5,7 @@ pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/access/AccessControlDefaultAdminRules.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "./IFeeModel.sol";
+import "./IReimbursementPool.sol";
 import "../lib/BLS12381.sol";
 import "../../threshold/IAccessControlApplication.sol";
 
@@ -68,6 +69,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
     uint16 public maxDkgSize;
     bool public isInitiationPublic;
     IFeeModel feeModel;  // TODO: Consider making feeModel specific to each ritual
+    IReimbursementPool reimbursementPool;
 
     constructor(
         IAccessControlApplication app,
@@ -131,6 +133,16 @@ contract Coordinator is AccessControlDefaultAdminRules {
         maxDkgSize = newSize;
     }
 
+    function setReimbursementPool(IReimbursementPool pool) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        if (address(pool) == address(0)){
+            delete reimbursementPool;
+        } else {
+            require(pool.isAuthorized(address(this)), "Coordinator not authorized");
+            reimbursementPool = pool;
+            // TODO: Events
+        }
+    }
+
     function numberOfRituals() external view returns(uint256) {
         return rituals.length;
     }
@@ -189,6 +201,8 @@ contract Coordinator is AccessControlDefaultAdminRules {
     }
 
     function postTranscript(uint32 ritualId, bytes calldata transcript) external {
+        uint256 initialGasLeft = gasleft();
+
         Ritual storage ritual = rituals[ritualId];
         require(
             getRitualState(ritual) == RitualState.AWAITING_TRANSCRIPTS,
@@ -219,6 +233,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
         if (ritual.totalTranscripts == ritual.dkgSize){
             emit StartAggregationRound(ritualId);
         }
+        processReimbursement(initialGasLeft);
     }
 
     function postAggregation(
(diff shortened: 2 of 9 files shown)


contracts/contracts/coordination/Coordinator.sol
```diff
@@ -45,26 +48,44 @@ contract Coordinator is Ownable {
...
+        uint16 _maxDkgSize,
+        address _admin,
+        address _feeModel
+    ) AccessControlDefaultAdminRules(0, _admin)
+    {
+        require(address(_feeModel.stakes()) == address(_stakes), "Invalid stakes for fee model");
+        application = _stakes;
         timeout = _timeout;
         maxDkgSize = _maxDkgSize;
+        feeModel = IFeeModel(_feeModel);
```
KPrasch, line 88: Shall we permit the `feeModel` address to be updated by the proper role?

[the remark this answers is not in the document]
KPrasch: Thanks for the reference 👍🏻  I guess my question is more about the type.  It's possible to call external contract functions from an `address` type?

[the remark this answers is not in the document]
cygnusv: Ah right, I'll change the constructor type. This PR is not ready, btw. Have not worked on tests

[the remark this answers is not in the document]
KPrasch: All good, just some stuff @theref and I noticed when working on #89 that's based over this PR, we didn't want to loose the thought.

cygnusv: Yeah, it's possible, but I'd rather handle that in a later PR, when we decide how do we want to manage fee models (see comment in L71). For our immediate needs (devnets), I don't foresee we will want to change the fee model

ape-config.yaml
```diff
@@ -23,13 +23,20 @@ solidity:
...
-    - "@openzeppelin/contracts=openzeppelin/v4.8.1"
-    - "@openzeppelin-upgradeable/contracts=openzeppelin-upgradeable/v4.8.1"
+    - "@openzeppelin/contracts=openzeppelin/v4.9.1"
+    - "@openzeppelin-upgradeable/contracts=openzeppelin-upgradeable/v4.9.1"
     - "@fx-portal/contracts=fx-portal/v1.0.5"
 
 deployments:
   polygon:
+    mainnet:
+      - contract_type: DAI
```
theref, line 33: 🚀

theref (approved):
This looks great

piotr-roslaniec (approved):

contracts/contracts/coordination/FlateRateFeeModel.sol
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    function getRitualInitiationCost(
+        address[] calldata providers,
+        uint32 duration
+    ) external view returns(uint256){
+        uint256 size = providers.length;
+        require(duration > 0, "Invalid ritual duration");
+        require(size > 0, "Invalid ritual size");
+        return feeRatePerSecond * size * duration;
```
arjunhassard, line 32: the proposal contains a default `duration` fixed at 182.5 days for initialization 
and then top ups are fixed at 30 days (payable every 30 days) to keep the availability horizon 182.5 days into the future

contracts/contracts/coordination/FlateRateFeeModel.sol
```diff
@@ -0,0 +1,34 @@
...
+    uint256 public immutable feeRatePerSecond;
+    IAccessControlApplication public immutable stakes;
+
+    constructor(IERC20 _currency, uint256 _feeRatePerSecond, address _stakes){
+        currency = _currency;
+        feeRatePerSecond = _feeRatePerSecond;
+        stakes = IAccessControlApplication(_stakes);
+    }
+
+    function getRitualInitiationCost(
```
arjunhassard, line 25: Since the 'paywall' here appears to be generating a public key for the first time, I'm wondering how a sponsor tops-up the availability horizon without a new ritual, and also if once they are conferred the`INITIATOR_ROLE`, they can do this unilaterally and whenever they want

arjunhassard (approved):
Huge step towards v7.0.0 👏

It may be that: 
(i) educating stakers on reimbursements 
(ii) adding it to staker UX 
(iii) maintaining & updating this section of the contracts 
may be more trouble than it's worth during genesis given that they will be 'paid back' by duration-based fees within a few days). 

Also, allowing permissionless ritual initiation is predicated on those rituals being far cheaper, so it's hard to know ex ante whether the reimbursements will be enough to justify the gas of withdrawal, and also exactly what the reimbursement could be bundled with and on what layer!

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -140,9 +186,11 @@ contract Coordinator is Ownable {
             newParticipant.provider = current;
             previous = current;
         }
+
+        processRitualPayment(id, providers, duration);
         
         // TODO: Include cohort fingerprint in StartRitual event?
-        emit StartRitual(id, msg.sender, providers);
+        emit StartRitual(id, ritual.authority, providers);
```
derekpierre, line 193: `StartRitual`'s 2nd parameter is indexed as "`initiator`":
- Should this be `ritual.initiator`?
- Should both `initiator` and `authority` be in the event?

derekpierre (commented):
Presumably we'll need respective `nucypher` changes (`CoordinatorAgent` etc.) before mergining into `main`.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -95,17 +116,31 @@ contract Coordinator is Ownable {
...
+    function setMaxDkgSize(uint16 newSize) external onlyRole(DEFAULT_ADMIN_ROLE) {
         emit MaxDkgSizeChanged(maxDkgSize, newSize);
         maxDkgSize = newSize;
     }
 
+    function setReimbursementPool(IReimbursementPool pool) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        if (address(pool) == address(0)){
+            delete reimbursementPool;
+        } else {
+            require(pool.isAuthorized(address(this)), "Coordinator not authorized");
```
vzotova, line 138: for me looks like it could be optimized slightly (unless I'm missing something):
`require(address(pool) == address(0) || pool.isAuthorized(address(this))`
instead of delete explicitly, just assign zero address. wdyt?

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -115,16 +150,27 @@ contract Coordinator is Ownable {
...
         uint256 length = providers.length;
         require(2 <= length && length <= maxDkgSize, "Invalid number of nodes");
+        require(duration > 0, "Invalid ritual duration");  // TODO: We probably want to restrict it more
 
         uint32 id = uint32(rituals.length);
         Ritual storage ritual = rituals.push();
-        ritual.initiator = msg.sender;  // TODO: Consider sponsor model
-        ritual.dkgSize = uint32(length);
+        ritual.initiator = msg.sender;
+        ritual.authority = authority;
```
vzotova, line 170: Could be `authority` zero address?

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -267,4 +326,55 @@ contract Coordinator is Ownable {
...
+        // Transfer fees back to initiator if failed
+        if(state == RitualState.TIMEOUT || state == RitualState.INVALID){
+            // Amount to refund depends on how much work nodes did for the ritual.
+            // TODO: Validate if this is enough to remove griefing attacks
+            uint256 executedTransactions = ritual.totalTranscripts + ritual.totalAggregations;
+            uint256 expectedTransactions = 2 * ritual.dkgSize;
+            uint256 consumedFee = pending * executedTransactions / expectedTransactions;
+            uint256 refundableFee = pending - consumedFee;
+            IERC20 currency = IERC20(feeModel.currency());
+            currency.transferFrom(address(this), ritual.initiator, refundableFee);
```
vzotova, line 366: Nitpick, but `safeTransferFrom` would be better

cygnusv: ✅

contracts/contracts/coordination/Coordinator.sol
cygnusv: Good catch. I meant `authority` here. IMO, once we introduce the authority role, the initiator is nothing more than a sponsor so there's in principle not much interest on it. In fact, once the fee is fully processed (so there's no refund), this role is not needed anymore. I would lean towards just including the authority in the event.

contracts/contracts/coordination/FlateRateFeeModel.sol
cygnusv: Is this something you would expect to be hardcoded at the smart contract level?

cygnusv: I'm opening an issue to track top-ups https://github.com/nucypher/nucypher-contracts/issues/93

contracts/contracts/coordination/FlateRateFeeModel.sol
cygnusv: Good questions. Let's track these here https://github.com/nucypher/nucypher-contracts/issues/93

KPrasch (approved):
Need a matching PR into nucypher/nucypher but looks good 👍🏻

contracts/contracts/coordination/Coordinator.sol
cygnusv: ✅

ape-config.yaml
cygnusv: Potentially. The deployment script for FlatRateFeeModel takes a  `currency` parameter address, but if you pass `DAI` it will look for the address here.

derekpierre (approved):
🎸

contracts/contracts/coordination/FlateRateFeeModel.sol
arjunhassard: No not hardcoded, as ofc we may need to change these. Just wanted to note the likely genesis defaults, pending feedback. However there is a small question of trust implications post-v7.0.0 – ideally the Threshold DAO would control sponsorship parameters like these

PR #89, opened 2023-06-21
TACo Encryption Allow Logic Interfaces

KPrasch:
Implements the contracts from the [specification](https://www.notion.so/WIP-Authorization-Specification-03f89660e721411c9fdfaef518a05fb5).

Co Authored By @theref @cygnusv @jMyles @derekpierre 

Based over #86

### DIFF
--- a/contracts/contracts/coordination/AllowList.sol
+++ b/contracts/contracts/coordination/AllowList.sol
@@ -1,22 +1,25 @@
 pragma solidity ^0.8.0;
-import "@openzeppelin/contracts/access/AccessControl.sol";
+import "@openzeppelin/contracts/access/AccessControlDefaultAdminRules.sol";
 import "./IAccessController.sol";
 import "./Coordinator.sol";
 
-contract AllowList is AccessControl, IAccessController {
+
+contract AllowList is AccessControlDefaultAdminRules, IAccessController {
     Coordinator public coordinator;
 
     // mapp
     mapping(uint256 => mapping(address => bool)) public rituals;
 
-    constructor(Coordinator _coordinator) {
+    constructor(
+        Coordinator _coordinator,
+        address _admin
+    ) AccessControlDefaultAdminRules(0, _admin) {
         coordinator = _coordinator;
     }
 
-    function bytesToAddress(bytes memory bys) private pure returns (address addr) {
-        assembly {
-            addr := mload(add(bys, 20))
-        }
+    function setCoordinator(Coordinator _coordinator) public {
+        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "Only admin can set coordinator");
+        coordinator = _coordinator;
     }
 
     function isEnricoAuthorized(
@@ -24,14 +27,14 @@ contract AllowList is AccessControl, IAccessController {
         bytes memory evidence,
         bytes memory ciphertextHash
     ) public view override returns(bool) {
-        enricoAddress = address(uint160(bytes20(evidence)));
+        address enricoAddress = address(uint160(bytes20(evidence)));
         return rituals[ritualID][enricoAddress];
     }
 
     function authorize(uint256 ritualID, address[] calldata addresses) public {
         require(coordinator.rituals(ritualId).authority == msg.sender,
             "Only ritual authority is permitted");
-        require(coordinator.getRitualStatus(ritualId) == RitualStatus.ACTIVE,
+        require(coordinator.getRitualStatus(ritualId) == RitualStatus.FINALIZED,
             "Only active rituals can add authorizations");
         for (uint i=0; i<addresses.length; i++) {
             rituals[ritualID][addresses[i]] = true;
@@ -41,7 +44,7 @@ contract AllowList is AccessControl, IAccessController {
     function deauthorize(uint256 ritualID, address[] calldata addresses) public {
         require(coordinator.rituals(ritualId).authority == msg.sender,
             "Only ritual authority is permitted");
-        require(coordinator.getRitualStatus(ritualId) == RitualStatus.ACTIVE,
+        require(coordinator.getRitualStatus(ritualId) == RitualStatus.FINALIZED,
             "Only active rituals can add authorizations");
         for (uint i=0; i<addresses.length; i++) {
             rituals[ritualID][addresses[i]] = false;
--- a/contracts/contracts/coordination/AllowList.sol
+++ b/contracts/contracts/coordination/AllowList.sol
@@ -1,0 +1,50 @@
+pragma solidity ^0.8.0;
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "./IAccessController.sol";
+import "./Coordinator.sol";
+
+contract AllowList is AccessControl, IAccessController {
+    Coordinator public coordinator;
+
+    // mapp
+    mapping(uint256 => mapping(address => bool)) public rituals;
+
+    constructor(Coordinator _coordinator) {
+        coordinator = _coordinator;
+    }
+
+    function bytesToAddress(bytes memory bys) private pure returns (address addr) {
+        assembly {
+            addr := mload(add(bys, 20))
+        }
+    }
+
+    function isEnricoAuthorized(
+        uint256 ritualID,
+        bytes memory evidence,
+        bytes memory ciphertextHash
+    ) public view override returns(bool) {
+        enricoAddress = address(uint160(bytes20(evidence)));
+        return rituals[ritualID][enricoAddress];
+    }
+
+    function authorize(uint256 ritualID, address[] calldata addresses) public {
+        require(coordinator.rituals(ritualId).authority == msg.sender,
+            "Only ritual authority is permitted");
+        require(coordinator.getRitualStatus(ritualId) == RitualStatus.ACTIVE,
+            "Only active rituals can add authorizations");
+        for (uint i=0; i<addresses.length; i++) {
+            rituals[ritualID][addresses[i]] = true;
+        }
+    }
+
+    function deauthorize(uint256 ritualID, address[] calldata addresses) public {
+        require(coordinator.rituals(ritualId).authority == msg.sender,
+            "Only ritual authority is permitted");
+        require(coordinator.getRitualStatus(ritualId) == RitualStatus.ACTIVE,
+            "Only active rituals can add authorizations");
+        for (uint i=0; i<addresses.length; i++) {
+            rituals[ritualID][addresses[i]] = false;
+        }
+    }
+}
--- a/contracts/contracts/coordination/AllowList.sol
+++ b/contracts/contracts/coordination/AllowList.sol
@@ -1,14 +1,13 @@
 pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/access/AccessControlDefaultAdminRules.sol";
-import "./IAccessController.sol";
+import "./IRitualAuthorizer.sol";
 import "./Coordinator.sol";
 
 
-contract AllowList is AccessControlDefaultAdminRules, IAccessController {
+contract AllowList is AccessControlDefaultAdminRules, IRitualAuthorizer {
     Coordinator public coordinator;
 
-    // mapp
-    mapping(uint256 => mapping(address => bool)) public rituals;
+    mapping(uint256 => mapping(address => bool)) public authorizations;
 
     constructor(
         Coordinator _coordinator,
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -9,6 +9,8 @@ import "./IFeeModel.sol";
 import "./IReimbursementPool.sol";
 import "../lib/BLS12381.sol";
 import "../../threshold/IAccessControlApplication.sol";
+import "../../../../nucypher/tests/acceptance/contracts/.cache/openzeppelin/v4.8.1/access/IAccessControl.sol";
+import "./IRitualAuthorizer.sol";
 
 /**
 * @title Coordinator
@@ -58,7 +60,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
         address authority;
         uint16 dkgSize;
         bool aggregationMismatch;
-        address accessController;
+        IRitualAuthorizer accessController;
         BLS12381.G1Point publicKey;
         bytes aggregatedTranscript;
         Participant[] participant;
@@ -92,7 +94,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
         uint16 _maxDkgSize,
         address _admin,
         IFeeModel _feeModel,
-        address _defaultAccessController
+        IRitualAuthorizer _defaultAccessController
     ) AccessControlDefaultAdminRules(0, _admin)
     {
         require(address(_feeModel.stakes()) == address(_stakes), "Invalid stakes for fee model");
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -58,6 +58,7 @@ contract Coordinator is AccessControlDefaultAdminRules {
         address authority;
         uint16 dkgSize;
         bool aggregationMismatch;
+        address accessController;
         BLS12381.G1Point publicKey;
         bytes aggregatedTranscript;
         Participant[] participant;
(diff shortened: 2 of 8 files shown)


contracts/contracts/coordination/AllowList.sol
```diff
@@ -0,0 +1,53 @@
...
+        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "Only admin can set coordinator");
+        coordinator = _coordinator;
+    }
+
+    function isEnricoAuthorized(
+        uint256 ritualID,
+        bytes memory evidence,
+        bytes memory ciphertextHash
+    ) public view override returns(bool) {
+        address enricoAddress = address(uint160(bytes20(evidence)));
```
cygnusv, line 30: An address can't be the evidence, otherwise anyone can put any address there. The evidence was defined as a signature, so what we want here is to do something like:
`address enricoAddress = ecrecover(ciphertextHash, v, r, s)`
where `v, r, s` are somehow extracted from the signature `evidence`, and `ciphertextHash` should be somehow casted to bytes32 (what `ecrecover` expects). See https://docs.soliditylang.org/en/latest/cheatsheet.html#mathematical-and-cryptographic-functions

cygnusv (commented):
Looking good! Some preliminar comments.

contracts/contracts/coordination/AllowList.sol
KPrasch: @theref and I considered recovering the address on-chain via ECDSA, however in the case of the `AllowList` for example, the only way an address can be "authorized" is by `Ritual.authority` setting it in a permissioned way.  In other word, only the ritual's authority can add new enricos.  Are we missing something?

contracts/contracts/coordination/AllowList.sol
```diff
@@ -0,0 +1,53 @@
...
+import "@openzeppelin/contracts/access/AccessControlDefaultAdminRules.sol";
+import "./IAccessController.sol";
+import "./Coordinator.sol";
+
+
+contract AllowList is AccessControlDefaultAdminRules, IAccessController {
+    Coordinator public coordinator;
+
+    // mapp
+    mapping(uint256 => mapping(address => bool)) public rituals;
```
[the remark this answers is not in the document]
KPrasch: ✔️

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -45,26 +48,48 @@ contract Coordinator is Ownable {
...
-        uint32 totalTranscripts;
-        uint32 totalAggregations;
-        BLS12381.G1Point publicKey;
+        uint32 endTimestamp;
+        uint16 totalTranscripts;
+        uint16 totalAggregations;
+        address authority;
+        uint16 dkgSize;
         bool aggregationMismatch;
+        address accessController;
```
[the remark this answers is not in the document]
KPrasch: ✔️

contracts/contracts/coordination/AllowList.sol
cygnusv: Let's say the authority authorizes `0x1234`. I don't own that address, but I just write `0x1234` as evidence when creating the conditions and access control policy. Ursulas would accept this evidence and that's not what we want. We need to provide `evidence` that I own that address.

contracts/contracts/coordination/IAccessControler.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.8.0;
+
+interface IAccessController {
```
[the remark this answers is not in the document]
KPrasch: 😆 ✔️   `IAccessController` -> `IRitualAuthorizer`

contracts/contracts/coordination/AllowList.sol
KPrasch: If enricos are submitting a signature of the cipertext and policy, doesn't that mitigate the issue you indicate (as enrico cannot simply "write `0x1234` as evidence"?  

Ursula will recover the address from the signature provided in `AccessControlPolicy.authorization` and check that it's in the allow list.

cygnusv: > If enricos are submitting a signature of the ciphertext and policy, doesn't that mitigate the issue you indicate (as enrico cannot simply "write `0x1234` as evidence")?

Yes, that's what I suggested initially.

> Ursula can recover the address from the signature provided in `AccessControlPolicy.authorization` and check that it's in the allow list.

Right, what I'm saying is that what you called `AccessControlPolicy.authorization` is what must be submitted as `evidence` by Ursula. That's what I originally meant by the evidence is attached to the ciphertext.

contracts/contracts/coordination/IAccessControler.sol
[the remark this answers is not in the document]
cygnusv: Is it RitualAuthorizer? The functionality here is to ascertain if the actor creating conditions/access control policies for a given ciphertext is authorized or not. This is related to the Enrico naming discussion, actually, because we depend on having a proper name for that. Maybe something like `PolicyAuthorController`? I don't know, this requires a wider discussion with everyone.

contracts/contracts/coordination/AllowList.sol
KPrasch: Hmmm, not sure I'm following - Is there a problem with trusting ursula to perform a valid signature recovery?  Is that different that trusting ursula to perform a valid condition evaluation?

KPrasch: Another way of framing it: What is the advantage of doing an authorization signature recovery onchian vs. in Ursula's runtime?

cygnusv: The main advantage I see is that this allows for modular and pluggable evidence evaluation mechanisms (via smart contracts), without having to change Ursula's code

theref:
@KPrasch I think you've overwritten some work here, I'll push to another branch so that we can compare

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
+pragma solidity ^0.8.0;
+import "@openzeppelin/contracts/access/AccessControlDefaultAdminRules.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "./IRitualAuthorizer.sol";
+import "./Coordinator.sol";
+
+contract GlobalAllowList is AccessControlDefaultAdminRules, IRitualAuthorizer {
+    using ECDSA for bytes32;
+
+    Coordinator public coordinator;
```
vzotova, line 10:
```suggestion
    Coordinator public immutable coordinator;
```

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+    mapping(uint256 => mapping(address => bool)) public authorizations;
+
+    constructor(
+        Coordinator _coordinator,
+        address _admin
+    ) AccessControlDefaultAdminRules(0, _admin) {
+        coordinator = _coordinator;
```
vzotova, line 17: Ideally add check that contract is valid. For example, call some view method in coordinator in `require`

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+          .recover(evidence);
+        return authorizations[ritualID][recovered_address];
+    }
+
+    function authorize(uint32 ritualID, address[] calldata addresses) public {
+        require(coordinator.getAuthority(ritualID) == msg.sender,
+            "Only ritual authority is permitted");
+        require(coordinator.isRitualFinalized(ritualID),
+            "Only active rituals can add authorizations");
+        for (uint i=0; i<addresses.length; i++) {
```
vzotova, line 41:
```suggestion
        for (uint256 i=0; i < addresses.length; i++) {
```

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+            authorizations[ritualID][addresses[i]] = true;
+        }
+    }
+
+    function deauthorize(uint32 ritualID, address[] calldata addresses) public {
+        require(coordinator.getAuthority(ritualID) == msg.sender,
+            "Only ritual authority is permitted");
+        require(coordinator.isRitualFinalized(ritualID),
+            "Only active rituals can add authorizations");
+        for (uint i=0; i<addresses.length; i++) {
```
vzotova, line 51:
```suggestion
        for (uint256 i=0; i < addresses.length; i++) {
```

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -184,11 +198,15 @@ contract Coordinator is AccessControlDefaultAdminRules {
         return ritual.participant;
     }
 
-    function initiateRitual(
+    function _initiateRitual(
```
derekpierre, line 201: Should we walk back this change i.e. change `_initiateRitual` -> `initiatRitual` and remove the function on L254, since the params are now the same?

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -359,13 +391,13 @@ contract Coordinator is AccessControlDefaultAdminRules {
     }
 
     function getParticipantFromProvider(
-        uint256 ritualID,
+        uint32 ritualID,
```
derekpierre, line 394: Should we be consistent with our camel-case (other spots as well - L400 and L412)?
```suggestion
        uint32 ritualId,
```

contracts/contracts/coordination/GlobalAllowList.sol
derekpierre: Can this variable be immutable - there is a `setCoordinator(...)` function?

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+          .toEthSignedMessageHash()
+          .recover(evidence);
+        return authorizations[ritualID][recovered_address];
+    }
+
+    function authorize(uint32 ritualID, address[] calldata addresses) public {
+        require(coordinator.getAuthority(ritualID) == msg.sender,
+            "Only ritual authority is permitted");
+        require(coordinator.isRitualFinalized(ritualID),
+            "Only active rituals can add authorizations");
```
derekpierre, line 40: Does it make sense to turn these requires into one of those solidity modifiers like `onlyOwner` etc., in this case `onlyAuthorityForActiveRitual` or something like that...

It can be used for authorizing and deauthorizing.

tests/test_coordinator.py
```diff
@@ -93,49 +93,93 @@ def coordinator(project, deployer, stake_info, flat_rate_fee_model, initiator):
     return contract
 
 
+@pytest.fixture()
+def global_allow_list(project, deployer, coordinator):
+    contract = project.GlobalAllowList.deploy(
```
derekpierre, line 98: Should we add tests (maybe not in this file) for the actual `GlobalAllowList` contract?

contracts/contracts/coordination/IRitualAuthorizer.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.8.0;
+
+interface IRitualAuthorizer {
```
derekpierre, line 3: 👍 Nice!

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+    ) AccessControlDefaultAdminRules(0, _admin) {
+        coordinator = _coordinator;
+    }
+
+    function setCoordinator(Coordinator _coordinator) public {
+        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "Only admin can set coordinator");
+        coordinator = _coordinator;
+    }
+
+    function isAuthorized(
```
piotr-roslaniec, line 25: Is the DKG client supposed to call this method at some point? Fore example, when picking nodes for the DKG cohort?

piotr-roslaniec (approved):

contracts/contracts/coordination/GlobalAllowList.sol
vzotova: you're right, thanks

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
+pragma solidity ^0.8.0;
+import "@openzeppelin/contracts/access/AccessControlDefaultAdminRules.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "./IRitualAuthorizer.sol";
+import "./Coordinator.sol";
+
+contract GlobalAllowList is AccessControlDefaultAdminRules, IRitualAuthorizer {
```
cygnusv, line 7: After re-reading this PR, I think a more descriptive name for the interface is `IEncryptionAuthorizer`, and subsequently the contract could be `GlobalEncryptionAllowList` (although not as necessary if the interface is renamed).

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+    constructor(
+        Coordinator _coordinator,
+        address _admin
+    ) AccessControlDefaultAdminRules(0, _admin) {
+        coordinator = _coordinator;
+    }
+
+    function setCoordinator(Coordinator _coordinator) public {
+        require(hasRole(DEFAULT_ADMIN_ROLE, msg.sender), "Only admin can set coordinator");
+        coordinator = _coordinator;
```
cygnusv, line 22: What's the use case for setting the Coordinator after deployment, if it was already set on construction time?

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    function isAuthorized(
+        uint32 ritualID,
+        bytes memory evidence,
+        bytes memory digest
+    ) public view override returns(bool) {
+        address recovered_address = keccak256(digest)
```
cygnusv, line 30: If `digest` is already a digest, there should be no need to call `keccak256`. Related to this is the lack of tests for this method. Maybe we can just put a pin on this for the moment and revisit in a separate PR, but something needs to be done here.

cygnusv (approved):
Great work! My main comment is regarding how hash within the authorization check is performed and the lack of tests for that.

contracts/contracts/coordination/Coordinator.sol
KPrasch: ✔️

contracts/contracts/coordination/Coordinator.sol
KPrasch: ✔️

contracts/contracts/coordination/GlobalAllowList.sol
KPrasch: ✔️

contracts/contracts/coordination/GlobalAllowList.sol
KPrasch: ✔️

contracts/contracts/coordination/GlobalAllowList.sol
KPrasch: In the event the coordinator is redeployed to support a security patch or new feature.

contracts/contracts/coordination/GlobalAllowList.sol
KPrasch: ✔️

contracts/contracts/coordination/GlobalAllowList.sol
KPrasch: ✔️

contracts/contracts/coordination/GlobalAllowList.sol
KPrasch: ✔️

KPrasch:
Thanks for all of the reviews and suggestions on this PR.  I introduced some new tests and covered every RFC.  Let me know if there's anything else before merging.  If you've already reviewed, perhaps just skim the new commits. 🤠

vzotova (approved):

derekpierre (approved):
🎸

PR #90, opened 2023-06-26
Support increased number of nodes in coordinator tests

piotr-roslaniec:
- Fixes `Providers not sorted` thrown in `test_coordinator.py` due to case-insensitive sorting of providers
- Based on #86 
- Need 2 reviews

### DIFF
--- a/ape-config.yaml
+++ b/ape-config.yaml
@@ -92,4 +92,4 @@ deployments:
 
 test:
   mnemonic: test test test test test test test test test test test junk
-  number_of_accounts: 15
+  number_of_accounts: 67
--- a/ape-config.yaml
+++ b/ape-config.yaml
@@ -92,4 +92,4 @@ deployments:
 
 test:
   mnemonic: test test test test test test test test test test test junk
-  number_of_accounts: 67
+  number_of_accounts: 10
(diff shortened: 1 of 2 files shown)


cygnusv:
> * Fixes `Providers not sorted` thrown in `test_coordinator.py` due to case-insensitive sorting of providers

Good catch!

> * Increases the number of test accounts to 67 (64 nodes + 3 technical accounts)

Not sure if increasing the number of nodes in the test is really necessary in the long-term. In this current shape, it's just going to make the overall test time to be longer. Our protocol, from the point of view of node operations to the Coordinator, is constant-time (constant-gas, if we're nit-picky) except for:
* Transcript size: This is what could really benefit from make it dependent on the cohort size, but we would need to have a clear formula for this (or at least, a good approximation)
* the function `getParticipantFromProvider()`, which I'm planning to make constant-time in a forthcoming PR

So, I'd propose to keep the test node count low (anything between 4 and 8 is fine) and make the transcript size depend on the node count.

piotr-roslaniec:
I agree with your analysis. I reverted the number of accounts and added a dynamic transcript size. For now, it uses an approximated formula and in the future, we can replace the actual `ferveo::Transcript` creation if needed.

manumonti (approved):
LGTM! 🙌

Maybe we should edit the description of this PR since AFAIK we are not increasing the number of test accounts to 67 with these changes. 

> - Increases the number of test accounts to 67 (64 nodes + 3 technical accounts)

PR #91, opened 2023-06-26
[WIP] Benchmark Coordinator methods gas usage

piotr-roslaniec:
- Based on #90
- See results [here](https://hackmd.io/dQ6aQF7ESLWrU9_2fh-swA)
- Not sure if we really need to merge this (in this form), but automated gas reporting sounds nice


cygnusv:
Very outdated and not relevant now. We can reopen this if needed.

PR #114, opened 2023-08-25
Operator confirmation from Coordinator to TACo application

vzotova:
Based on #102
Fixes #112  
Refs #69

* Renames `StakeInfo` and related interfaces
* Changes node confirmation: `Coordinator.setPublicKey` is the source of node confirmations

PR #115, opened 2023-09-05
TACo commitment

vzotova:
Based on #114 
Now you can make a commitment to TACo!
* 4 options, some of them can be empty
* can't request decrease until end of commitment
* can be done only once


contracts/contracts/TACoApplication.sol
```diff
@@ -440,6 +467,10 @@ contract TACoApplication is IApplication, ITACoChildToRoot, OwnableUpgradeable {
             _toAmount == 0 || _toAmount >= minimumAuthorization,
             "Resulting authorization will be less than minimum"
         );
+        require(
+            info.endCommitment <= block.timestamp,
+            "Can't request deauthorization before end of commitment"
+        );
```
cygnusv, line 473: For readers, this is where the magic happens.

cygnusv (approved):

cygnusv (approved):

PR #116, opened 2023-09-07
[WIP] Tweaks for Alpha 12

derekpierre:
Associated changes based on testing from https://github.com/nucypher/nucypher/pull/3213 work.

### DIFF
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -163,9 +163,9 @@ contract Coordinator is AccessControlDefaultAdminRules, FlatRateFeeModel {
     ) external view returns (BLS12381.G2Point memory) {
         ParticipantKey[] storage participantHistory = keysHistory[_provider];
 
-        for (uint256 i = participantHistory.length - 1; i >= 0; i--) {
-            if (participantHistory[i].lastRitualId <= _ritualId) {
-                return participantHistory[i].publicKey;
+        for (uint256 i = participantHistory.length; i > 0; i--) {
+            if (participantHistory[i - 1].lastRitualId <= _ritualId) {
+                return participantHistory[i - 1].publicKey;
             }
         }
 


derekpierre:
FYI @vzotova , @cygnusv - this is what @KPrasch and I are currently testing with for https://github.com/nucypher/nucypher/pull/3213 as the first stage. These commits are based on the commit before @vzotova 's recent TACoRoot/Child Application changes.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -401,17 +407,26 @@ contract Coordinator is AccessControlDefaultAdminRules, FlatRateFeeModel {
...
     }
 
+    function isEncryptionAuthorized(
+        uint32 ritualId,
+        bytes memory evidence,
+        bytes memory ciphertextHeader
+    ) external view returns (bool) {
+        Ritual storage ritual = rituals[ritualId];
+        require(getRitualState(ritual) == RitualState.FINALIZED, "Ritual not finalized");
+        return ritual.accessController.isAuthorized(ritualId, evidence, ciphertextHeader);
```
derekpierre, line 417: Thinking a bit more about this - @vzotova , @cygnusv are there any security risks to having the Coordinator contract call some random access controller contract? (cc @KPrasch)

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -401,17 +407,26 @@ contract Coordinator is AccessControlDefaultAdminRules, FlatRateFeeModel {
         return getParticipantFromProvider(rituals[ritualId], provider);
     }
 
+    function isEncryptionAuthorized(
```
piotr-roslaniec, line 410: Should the client call this method before attempting to encrypt? Or is it only meant to be called by Ursulas?

piotr-roslaniec (approved):

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -56,6 +56,7 @@ contract Coordinator is AccessControlDefaultAdminRules, FlatRateFeeModel {
     }
 
     struct Ritual {
+        // NOTE: changing the order here affects nucypher/nucypher: CoordinatorAgent
```
cygnusv, line 59: Good call!

contracts/contracts/coordination/Coordinator.sol
cygnusv: Not a problem since all the call chain is made of `view` methods.

cygnusv (approved):

contracts/contracts/coordination/Coordinator.sol
KPrasch: It is meant to be called by Ursula.  This is convenient so that Ursula does not need to handle call forwarding and interface ABIs.

cygnusv: Just for Ursulas, particularly since one of the parameters is the ciphertext digest, so the encryptor could only call it after the fact. But you're comment is a good one; the encryptor can call this method in the GlobalAllowList contract:
https://github.com/nucypher/nucypher-contracts/blob/74ac5eaf14a40e57652072531eae7c7f90bcfa1a/contracts/contracts/coordination/GlobalAllowList.sol#L38-L40

Note that this method is specific to `GlobalAllowList`, and not to the `IEncryptionAuthorizer` interface. I think we probably need something there as well.

cygnusv: Opened #118

contracts/contracts/coordination/Coordinator.sol
derekpierre: 👍

KPrasch (approved):

PR #126, opened 2023-10-02
Implements multi-chain JSON registry data formatting

KPrasch:
##### Requested Reviews
2

##### What this does

- Implements multi-chain JSON registry data format introduced in https://github.com/nucypher/nucypher/pull/3261
- Enriches type annotation and aliasing in `deployment/registry.py`
- Modifies `merge_registries` to include chain ID as part of conflict assessment criteria
- Optimizes registry merge utilities

### DIFF
--- a/deployment/registry.py
+++ b/deployment/registry.py
@@ -1,36 +1,43 @@
 import json
+from collections import defaultdict
 from enum import Enum
 from pathlib import Path
-from typing import Dict, List, NamedTuple, Optional
+from typing import Dict, List, NamedTuple, Optional, Union
 
 from ape.contracts import ContractInstance
 from eth_typing import ChecksumAddress
 from eth_utils import to_checksum_address
+from web3.types import ABI
 
 from deployment.params import get_contract_container
 from deployment.utils import check_registry_filepath
 
 
-class RegistryEntry(NamedTuple):
+ChainId = int
+ContractName = str
+ContractArtifacts = Dict[ContractName, Union[ChecksumAddress, ABI]]
+RegistryData = Dict[ChainId, ContractArtifacts]
+
+
+class ContractEntry(NamedTuple):
     """Represents a single entry in a nucypher-style contract registry."""
+    chain_id: ChainId
+    name: ContractName
+    address: ChecksumAddress
+    abi: ABI
 
-    contract_name: str
-    contract_version: str  # TODO: remove version from registry
-    contract_address: ChecksumAddress
-    contract_abi: List[Dict]
 
-
-def _get_abi(contract_instance: ContractInstance) -> List[Dict]:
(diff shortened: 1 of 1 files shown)


deployment/registry.py
```diff
@@ -114,89 +125,82 @@ def _select_conflict_resolution(
...
+    # Read the registries, excluding deprecated contracts
+    reg1 = {e.name: e for e in read_registry(registry_1_filepath) if e.name not in deprecated_contracts}
+    reg2 = {e.name: e for e in read_registry(registry_2_filepath) if e.name not in deprecated_contracts}
+
+    merged = defaultdict(dict)
+
+    # Iterate over all unique contract names across both registries
+    for name in set(reg1) | set(reg2):
+        entry = reg1.get(name) or reg2.get(name)
+        conflict_entry = reg2.get(name) if name in reg1 else None
```
derekpierre, line 170: Nice!

deployment/registry.py
```diff
@@ -1,36 +1,43 @@
...
+ContractName = str
+ContractArtifacts = Dict[ContractName, Union[ChecksumAddress, ABI]]
+RegistryData = Dict[ChainId, ContractArtifacts]
+
 
-    contract_name: str
-    contract_version: str  # TODO: remove version from registry
-    contract_address: ChecksumAddress
-    contract_abi: List[Dict]
+class ContractEntry(NamedTuple):
```
derekpierre, line 22: What's the reason we use two different representations for the same information i.e. `RegistryData` and `ContractEntry`

KPrasch: `ContractEntry` is a single contract entry in the registry abstracted into a `NamedTuple` used for internal processing.  `RegistryData` is the entire registry data structure as a `Dict` type alias for type annotations.

derekpierre: I didn't express this very well. 

Basically, `RegistryData` only applies to writing to the registry, vs `ContractEntry`, which is used everywhere else. 

Perhaps `write_registry(...)` should take List[ContractEntry] as the parameter, and generate the `RegistryData` internally instead of having callers of the function (`merge_registry`, `registry_from_ape_deployments`) do the conversion to RegistryData before calling. That way there is only place doing the conversion to `RegistryData`...and you barely even really need a separate type because it is no longer a parameter.

KPrasch: Nice suggestions ✔️

piotr-roslaniec (approved):

derekpierre (approved):
🎸

cygnusv (approved):
LGTM! Just a minor comment

deployment/registry.py
derekpierre: ❤️  - 👨‍🍳 😗

KPrasch: ✔️

PR #131, opened 2023-10-03
Yaml Deployment Parameters

KPrasch:
- Use YAML instead of JSON for deployment parameter expression
- Follow-up PR #132


deployment/constructor_params/lynx/root.yml
```diff
@@ -0,0 +1,27 @@
+deployment:
```
derekpierre, line 1: This looks pretty nice - nicer than json, that's for sure.

derekpierre (approved):
🎸

deployment/params.py
```diff
@@ -1,14 +1,19 @@
...
 
+import yaml
 from ape import chain, project
 from ape.api import AccountAPI, ReceiptAPI
 from ape.cli import get_user_selected_account
 from ape.contracts.base import ContractContainer, ContractInstance, ContractTransactionHandler
 from ape.utils import ZERO_ADDRESS
+from eth_typing import ChecksumAddress
+from ethpm_types import MethodABI
+from web3.auto.gethdev import w3
```
KPrasch, line 14: Unrelated to this PR, but I'm wondering why we're using gethdev here instead of instantiating a headless `w3 = Web3()` instance. 🤔 Probably just an auto-importing oversight.

derekpierre: I grabbed the code from here - https://web3py.readthedocs.io/en/stable/web3.main.html#check-encodability.

PR #133, opened 2023-10-04
Expand Registry Artifacts to include Deployment Info

KPrasch:
Includes deployment transaction hash, block number, and deployer account in registry entries.

Example Artifact:
```json
{
    "1337": {
        "TACoApplication": {
            "address": "0x1DAcA792255f6D727Ba28306b4239073C060eCa1",
            "abi": [...],
            "tx_hash": "0x6b5703200e4d1f50290707fb850f72fc066291f87347cbb8f4790684e33525f0",
            "block_number": 5,
            "deployer": "0x2192f6112a026bce4047CeD2A16553Fd31E798B6"
        },
    ...
}
```

### DIFF
--- a/deployment/registry.py
+++ b/deployment/registry.py
@@ -23,6 +23,9 @@ class RegistryEntry(NamedTuple):
     name: ContractName
     address: ChecksumAddress
     abi: ABI
+    tx_hash: str
+    block_number: int
+    deployer: str
 
 
 def _get_abi(contract_instance: ContractInstance) -> ABI:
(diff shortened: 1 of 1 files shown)


derekpierre:
Haven't looked at the diff as yet, but one thing that stands out based on the description is that merging older registries may be difficult. Remember that we need to merge the registry entry for `SubscriptionManager` from the old registry files.

Something we need to think about - perhaps #126 also causes this issue... (cc @cygnusv )

derekpierre:
Of course we could also have a special case for the older registry since we only need to do it once.

KPrasch:
> Of course we could also have a special case for the older registry since we only need to do it once.

[#3261](https://github.com/nucypher/nucypher/pull/3261) migrates the data format of existing registries appropriately, but in general manual ad-hoc conversion is acceptable for any missing fields.

piotr-roslaniec (approved):

derekpierre (approved):
🎸

PR #136, opened 2023-10-06
Converts Mainnet Artifacts

KPrasch:
- utility to convert legacy NPM artifacts 
- converts mainnet artifacts
- based over #135


manumonti (approved):
Approved! LGTM

Issue #140, opened 2023-10-06
Clean up TapirSet and LynxSet to use common code


derekpierre:
- MockPolygonRoot/Child is pretty general but only sits in LynxSet.sol at the moment
- For Tapir deployment we just copied over work, but it can be cleaned up.

PR #141, opened 2023-10-06
Tapir Deployment

KPrasch:
### What this does
- Introduces `TapirSet.sol` (derived from `LynxSet.sol`) (see follow-up #140)
- Introduces Constructor parameters and deployment scripts for Tapir
- Includes Tapir deployment artifacts for Sepolia and Polygon/Mumbai

##### Discoveries
-  Always use `verify=False` to prevent crashing mid-deployment due to failed block explorer verification (see #142)
- Clear the ape deployment cache before running deployment scripts (`~/.ape/etherum|polygon`) (see #143)
- It is better to use a native desktop terminal window as opposed to an integrated IDE terminal to avoid visual artifacts that prevent the user from being able to see the console input line

##### Notes
Console output from deployment https://gist.github.com/KPrasch/4fc07f74295ba3e658e29d59e1c657a9


derekpierre (approved):
🎸

PR #145, opened 2023-10-06
TACo application refinements

vzotova:


contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -83,16 +83,20 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
...
         address oldOperator = info.operator;
 
-        if (stakingProvider != address(0) && operator != oldOperator) {
-            info.operator = operator;
-            // Update operator to provider mapping
-            stakingProviderFromOperator[oldOperator] = address(0);
-            stakingProviderFromOperator[operator] = stakingProvider;
-            info.operatorConfirmed = false;
-            // TODO placeholder to notify Coordinator
+        if (stakingProvider == address(0) || operator == oldOperator) {
```
manumonti, line 86: Sorry if it's a dumb question, but why not manage this with a `require` statement?

```suggestion
require (stakingProvider != address(0) && operator != oldOperator);
```

manumonti (approved):
LGTM ✌️! I left a question more than a suggestion.

contracts/contracts/coordination/TACoChildApplication.sol
vzotova: it's not dumb question. Because there is no real point to fail in xchain txs, it will be just skipped and that's it. But some sanity check to decrease gas consumption we want to do

manumonti: got it, thanks!

contracts/contracts/TACoApplication.sol
```diff
@@ -143,10 +144,13 @@ contract TACoApplication is IApplication, ITACoChildToRoot, OwnableUpgradeable {
         uint96 deauthorizing; // TODO real usage only in getActiveStakingProviders, maybe remove?
         uint64 endDeauthorization;
         uint96 tReward;
-        uint96 rewardPerTokenPaid;
+        uint160 rewardPerTokenPaid;
         uint64 endCommitment;
     }
 
+    uint256 public constant REWARD_PER_TOKEN_MULTIPLIER = 10 ** 4;
```
derekpierre, line 151: What does this constant represent? What's the multiplier? Is this something for precision?

deployment/constructor_params/lynx/child.yml
```diff
@@ -3,27 +3,28 @@ deployment:
...
-        _currency: $LynxRitualToken
-        _feeRatePerSecond: 1
-    - GlobalAllowList:
-        _coordinator: $Coordinator
-        _admin: $deployer
+  - MockPolygonChild
+  - ProxyAdmin
+  - LynxTACoChildApplication:
+      _rootApplication: $MockPolygonChild
+      _minimumAuthorization: 40000000000000000000000
```
derekpierre, line 14: Something to think about here is that the value used in `root.yml` for the root application would need to match the value used here... 🤔  (cc @KPrasch )

cygnusv: That's why we need a separate constants section or file

derekpierre: Indeed. Now that it is a yml file, perhaps we can specify a dependent constants.yml file perhaps...I'm less familiar with yml spec.

contracts/contracts/TACoApplication.sol
vzotova: yeah, name is not perfect. Thing is we have variable called `rewardPerToken`. With increase of precision now it's actually not per token but per 10000 tokens (that multiplier).

cygnusv: `REWARD_CALCULATION_MULTIPLIER`?

vzotova: hmm? this name suits better for `FLOATING_POINT_DIVISOR`

KPrasch (approved):

theref (approved):

PR #153, opened 2023-10-12
authorize enrico script

KPrasch:
```
$ ape run authorize_enrico --network polygon:mumbai:infura -r ./deployment/artifacts/lynx.json

Checking plugins...

0. testnet-deployer
Or 'TEST::account_idx', where `account_idx` is in [0..{len_test_accounts}]

Select an account: 0
Enter ritual ID: 1
Enter address to authorize: 0x....

Transacting GlobalAllowList[0x7b521E78].authorize(uint32 ritualId, address[] addresses) with arguments:
	ritualId=1
	addresses=['...']
Continue Y/N? y

DynamicFeeTransaction:
  chainId: 80001
  to: ...
  from: ...
  gas: 61159
  nonce: 322
  value: 0
  data: 0x75ab31...0d4cd7
  type: 2
  maxFeePerGas: 2425000030
  maxPriorityFeePerGas: 2425000000
  accessList: []

Sign:  [y/N]: y
Enter passphrase to unlock 'testnet-deployer' []: 
Leave 'testnet-deployer' unlocked? [y/N]: y
Confirmations (1/1): 100%|████████████████████████████████████████████████████████████| 1/1 [00:04<00:00,  4.80s/it]
INFO: Confirmed 0x... (total fees paid = 148310575917385)
```


scripts/lynx/authorize_enrico.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/python3
```
derekpierre, line 1: What do you think above moving this script to the common set of scripts `scripts/` and use click options? We're probably going to have to do something similar for `tapir` / `mainnet`.

Could do something like
```
ape run authorize_enrico --network polygon:mumbai:infura --registry-filepath <> --enrico <>
```

KPrasch: sounds like a great idea 👍🏻 we're gonna need documentation 😅

KPrasch: opened https://github.com/nucypher/nucypher-contracts/issues/154

theref (approved):

scripts/lynx/authorize_enrico.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def main():
+    check_plugins()
+    transactor = Transactor()
+    deployments = contracts_from_registry(filepath=REGISTRY_FILEPATH)
+    global_allow_list = deployments[project.GlobalAllowList.contract_type.name]
+    ritual_id = int(input("Enter ritual ID: "))
+    addresses = [to_checksum_address(input("Enter address to authorize: "))]
+    transactor.transact(global_allow_list.authorize, ritual_id, addresses)
```
derekpierre, line 22: Maybe this could be done as part of #154 as well, but should we check global_allow_list.isAddressAuthorized(...) first to potentially save some gas in case enrico is already authorized. 

Similar to what we do here - https://github.com/nucypher/nucypher/blob/development/scripts/hooks/nucypher_dkg.py#L336.

Just something to think about.

derekpierre (approved):
🎸

scripts/lynx/authorize_enrico.py
KPrasch: ✔️

PR #156, opened 2023-10-16
chore: remove lockfile

piotr-roslaniec:
- Just one review
- Update: Ouch, it doesn't work.


piotr-roslaniec:
Redundant because of https://github.com/nucypher/nucypher-contracts/issues/161

PR #160, opened 2023-10-20
Oz upgrade

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Upgrades OpenZeppelin library to 5.0.0

**Issues fixed/closed:**
Fixes #159 

**Why it's needed:**
It will be hard to upgrade to 5.x.x version after mainnet

**Notes for reviewers:**
Based on #145

PR #162, opened 2023-10-25
`authorizationParameters` and Solidity -> 0.8.22

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Adds `TACoApplication.authorizationParameters` and updates solidity version

**Why it's needed:**
Help with dashboard
Solidity version improves gas consumption

Based on #160


contracts/contracts/TACoApplication.sol
```diff
@@ -277,6 +286,39 @@ contract TACoApplication is IApplication, ITACoChildToRoot, OwnableUpgradeable {
...
+    ///         earning rewards and not being active in the network.
+    /// @return authorizationDecreaseChangePeriod Authorization decrease change
+    ///        period in seconds. It is the time, before authorization decrease
+    ///        delay end, during which the pending authorization decrease
+    ///        request can be overwritten.
+    ///        If set to 0, pending authorization decrease request can not be
+    ///        overwritten until the entire `authorizationDecreaseDelay` ends.
+    ///        If set to value equal `authorizationDecreaseDelay`, request can
+    ///        always be overwritten.
+    function authorizationParameters()
```
theref, line 306: woo!! 🚀 Thank you @vzotova

theref (approved):

KPrasch (approved):

tests/application/conftest.py
```diff
@@ -27,13 +26,11 @@
 
 REWARD_DURATION = 60 * 60 * 24 * 7  # one week in seconds
 DEAUTHORIZATION_DURATION = 60 * 60 * 24 * 60  # 60 days in seconds
-TOTAL_SUPPLY = Web3.to_wei(1_000_000_000, "ether")  # TODO NU(1_000_000_000, 'NU').to_units()
+TOTAL_SUPPLY = Web3.to_wei(10_000_000_000, "ether")  # TODO NU(10_000_000_000, 'NU').to_units()
```
derekpierre, line 29: If defining here then remove L25 - that's what influenced the testing confusion.

derekpierre: Side note, there is going to be a mint before TACo launch that could push the total supply closer to 11B, perhaps we should use that value.

contracts/contracts/TACoApplication.sol
```diff
@@ -148,7 +147,7 @@ contract TACoApplication is IApplication, ITACoChildToRoot, OwnableUpgradeable {
         uint64 endCommitment;
     }
 
-    uint256 public constant REWARD_PER_TOKEN_MULTIPLIER = 10 ** 4;
+    uint256 public constant REWARD_PER_TOKEN_MULTIPLIER = 10 ** 3;
```
derekpierre, line 150: This was some recommendation from the audit right, so instead of 10,000 we are doing 1000. Is that presumably still in line with what the original concern was in the audit?

tests/application/conftest.py
vzotova: 11B comparing to 10B is not important because main thing is to catch order of values

contracts/contracts/TACoApplication.sol
vzotova: they suggested `uint256` and for `uint256` it could be 10^9, for `uint160` smaller and I tried to squeeze max because I still doubt overflow will ever be possible

tests/application/conftest.py
vzotova: ✔️

derekpierre (approved):
🎸

PR #163, opened 2023-10-25
Add support for deployment constants

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


deployment/constructor_params/tapir/root.yml
```diff
@@ -22,6 +22,6 @@ contracts:
   - TransparentUpgradeableProxy:
       _logic: $TACoApplication
       admin_: $ProxyAdmin
-      _data: $bytes:0x
+      _data: 0x
```
KPrasch, line 25: 💥

deployment/params.py
```diff
@@ -242,66 +122,156 @@ def _validate_transaction_args(args: typing.Tuple[Any, ...], abi) -> typing.Dict
...
     """Represents the constructor parameters for a set of contracts."""
 
     class Invalid(Exception):
         """Raised when the constructor parameters are invalid"""
 
-    def __init__(self, parameters: OrderedDict):
-        validate_constructor_parameters(parameters)
+    def __init__(self, parameters: OrderedDict, constants: dict = None):
         self.parameters = parameters
+        self.constants = constants or dict()
```
KPrasch, line 133: 👍🏻

deployment/params.py
```diff
@@ -242,66 +122,156 @@ def _validate_transaction_args(args: typing.Tuple[Any, ...], abi) -> typing.Dict
...
+
+            # validate value type
+            value_to_validate = value
+            if _is_variable(value):
+                value_to_validate = self._resolve_param(value)
+            if not w3.is_encodable(abi_input.type, value_to_validate):
+                raise ConstructorParameters.Invalid(
+                    f"Constructor param name '{name}' at position {position} has a value '{value}' "
+                    f"whose type does not match expected ABI type '{abi_input.type}'"
+                )
```
KPrasch, line 274: many of (all?) of these methods are stateless and do not need access to the class's state -- consider relocating them to a helper module of pure functions to keep this class small, improve modularity and unit testing ease.

cygnusv: Not really. Once I created `_resolve_constants` to be stateful (as it uses constants defined in the config), then practically all the methods refactored here had dependencies going up. I tried to move as less things as possible.

KPrasch: The trade off of injecting `self.constants[name]` into `_resolve_constants` as a parameter input to decouple all these methods from the class is a good one imo.

cygnusv: `_resolve_constants` => `_resolve_special_variable` => `_resolve_param` => `_validate_constructor_abi_inputs`
 => `validate_constructor_parameters`

KPrasch: Precisely -- if you inject into `self.constants[name]` into `_resolve_constants` instead of using instance state, the rest of the methods can be relocated as helper functions too.

cygnusv: Addressed in 9b23df2b1e1be0870a4e8a2f4e93d4b68a384a1d (Thanks @KPrasch 🙏 )

KPrasch (approved):
LGTM 🤠 useful feature with a straightforward and precise implementation

deployment/params.py
```diff
@@ -105,6 +101,18 @@ def _resolve_deployer() -> str:
         return deployer_account.address
 
 
+def _validate_transaction_args(args: typing.Tuple[Any, ...], abi) -> typing.Dict[str, Any]:
+    """Validates the transaction arguments against the function ABI."""
+    named_args = dict()
+    for arg, abi_input in zip(args, abi.inputs):
```
derekpierre, line 107: Should we double-check that the length of `args` and `abi.inputs` are the same? Does `zip` fail if the lengths are different, or does it drop values?

KPrasch: Fair point... unfortunately by default `zip` will silently carry on even if the inputs are different lengths...

There is a way to enable this check In python 3.10+: 

> [PEP 618](https://www.python.org/dev/peps/pep-0618): The [zip()](https://docs.python.org/3.10/library/functions.html#zip) function now has an optional strict flag, used to require that all the iterables have an equal length.

```suggestion
    for arg, abi_input in zip(args, abi.inputs, strict=True):
```

derekpierre: 👍 If we don't want to worry about python version, we could just check the `len(args) == len(abi.inputs)`. We do something similar in `_validate_constructor_abi_inputs(...)`.

derekpierre:
@cygnusv are constants actually used in the configuration yml as part of this PR? I thought we were hoping to use it for something like dkg timeout etc.?

deployment/params.py
KPrasch: tbh it will be great to just support the latest python in this repo since we have the flexibility

derekpierre: Yep, I'm fine either way - just giving options.

KPrasch:
This PR now needs a rebase to resolve conflicts following the merge of #160 and #162

deployment/params.py
cygnusv: While working in adding the suggested check I realized 3 things:
1. The check is redundant because the prior calling this function, the `_get_function_abi` called just before that gives you the abi that has exactly the required number of arguments.
2. The current approach is brittle as Solidity has function polymorphism, i.e., many functions can share the same name, as long as they have different parameter types (or number of parameters), and we're assuming that we can make a distinction based solely on number of parameters.
3. While I'm fixing 1 and 2 in an additional commit to this PR (fe592c3a37a174cc1837274a5bc5fab758e07017), I think we're very likely over-validating stuff that web3py and ape already do, and consequently, unnecessarily increasing code complexity. Not for this PR, but we probably should subtract a little.

derekpierre: The underlying issue is that we try to do eager validation **before** executing a contract call (eager validation). Web3py is great for validation when actually calling a contract function, but not so much on the eager side. We do the eager validation on purpose because our scripts do multiple deployments/actions, and having it fail in the middle during the contract call is annoying, and potentially problematic, so we try the eager check to reduce such simplistic errors.

That being said, I don't doubt that we can better leverage web3py than we do currently, perhaps reuse common validation calls that web3 py uses to reduce our own logic - something like https://github.com/ethereum/web3.py/blob/main/web3/_utils/contracts.py#L144. Subtracting a little would be good.

deployment/constructor_params/lynx/child.yml
```diff
@@ -6,20 +6,25 @@ artifacts:
...
+   ONE_HOUR_IN_SECONDS: 3600
+   FORTY_THOUSAND_TOKENS_IN_WEI_UNITS: 40000000000000000000000  
+   TEN_MILLION_TOKENS_IN_WEI_UNITS: 10000000000000000000000000  # https://www.youtube.com/watch?v=EJR1H5tf5wE
+
 contracts:
   - MockPolygonChild
   - LynxTACoChildApplication:
       _rootApplication: $MockPolygonChild
-      _minimumAuthorization: 40000000000000000000000
+      _minimumAuthorization: $FORTY_THOUSAND_TOKENS_IN_WEI_UNITS
```
derekpierre, line 18: Constants look great - I like the `$` prefix.

PR #170, opened 2023-10-31
Multi-proxy deployments

### LINKED ISSUE #169: Support multiple proxies in one constructor params yml

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Add support for multiple proxy deployments in one constructor params yml

## Defining Proxies
Proxies can be defined as follows in the following ways:

1. Just "proxy:"
```yml
  - TACoApplication:
      proxy:
      constructor:
        _token: $LynxStakingToken
        ...
```
This causes the contract to be proxied (using `TransparentUpgradeableProxy` with default values:
- `_logic` - implicitly the contained contract; `TACoApplication` in this case.
- `initialOwner` = "$deployer" (special variable for the current deployer account be used)
- `_data` = 0x

2. With a `contract_type` defined:
```yml
  - LynxTACoChildApplication:
      proxy:
        contract_type: TACoChildApplication
      constructor:
        _rootApplication: $MockPolygonChild
        _minimumAuthorization: $FORTY_THOUSAND_TOKENS_IN_WEI_UNITS
```
This causes the contract to be proxied the same as 1) but the contract type for the proxy is the specified type; `TACoChildApplication` in this case. This is used for ABI for contract instance, and uses the contract_type name in the contract registry. It's for the case above where the custom contract is written, but should be presented as a different contract type

3. With overridden parameters
```yml
  - LynxTACoChildApplication:
      proxy:
        contract_type: TACoChildApplication
        constructor:
          initialOwner: "0xdeadbeef."
      constructor:
        _rootApplication: $MockPolygonChild
        _minimumAuthorization: $FORTY_THOUSAND_TOKENS_IN_WEI_UNITS
```
In this case all default parameter values are used for the proxy except for `initialOwner` which will use the "0xdeadbeef" value. Basically specific parameters can be overridden when defined in the yml, without needing to specify all parameters for the proxy constructor.

*NOTE:* `_logic` parameter cannot be overridden since it is implicitly the contract address for the contract the proxy definition is located in. An exception will raise of `_logic` is attempted to be defined.

### Overriding `_data` to do an initialization call on underlying logic contract
You can provide the call data to be encoded for `_data` value of the Proxy via the `$encode` variable for encoding functionality:

Examples:
- `$encode:initialize` - generate bytes for `_data` to call empty initialize function on logic contract during proxy initialization
- `$encode:myFunc,arg1,arg2,arg3` - generate bytes for `_data` to call `myFunc(arg1,arg2,arg3)` on logic contract during proxy initialization

This also means that for some deployment scripts, they no longer need to call `initialize` (TACoApplication case) in the python script, since providing the data for the in the yml will automatically take care of it.

## Using proxies
In the yml, you no longer need to use `$TransparentUpgradeableProxy:<ContractName>` to refer to the proxied version of the intended contract, now you can just refer to the contract as a simple contract variable $<ContractName>. For example to refer to the proxied LynxTACoChildApplication from example 3) above you can do the following:
```yml
  - Coordinator:
      constructor:
        _application: $LynxTACoChildApplication
        _timeout: $ONE_HOUR_IN_SECONDS
        _maxDkgSize: 4
        _admin: $deployer
        _currency: $LynxRitualToken
        _feeRatePerSecond: 1
```

## Deployment Scripts
Deployment scripts no longer need to do anything related to proxies, the definition in the yml is sufficient. Usage in the deployment script is as simple as (from example 3)):
```
taco_child_application = deployer.deploy(project.LynxTACoChildApplication)
```

The returned `taco_child_application` contract instance is the proxied contract for the `LynxTACoChildApplication` contract.

## Try it out

Change the name of the artifacts filename in the relevant [lynx/tapir]/[root/child].yml file:
```yml
artifacts:
    dir: ./deployment/artifacts/
    filename: test-<root or child>-<domain>.json
```

This is needed because we don't allow overwriting of conflicting artifacts files.

And then  try out the scripts using:
- Lynx
```bash
 $ ape run lynx deploy_child --network ethereum:local:test
 ...
  $ ape run lynx deploy_root --network ethereum:local:test
```

- Tapir
```bash
 $ ape run tapir deploy_child --network ethereum:local:test
 ...
  $ ape run tapir deploy_root --network ethereum:local:test
```

**Issues fixed/closed:**
> - Fixes #...

Fixes #169 

**Why it's needed:**

**Notes for reviewers:**


derekpierre:
There was an idea from @cygnusv that instead of using "constructor:" and "proxy:" to separate constructor parameters and proxy parmeters, use some sort of prefix (eg `__` or `$` although perhaps use something other `$` since we use `$` for variables) to separate additional macro type functionality (eg. proxy) from constructor parameters. It would allow contracts that don't care about proxy to not have to use "constructor:" definition. For example:

Instead of defining currently as:
```yml
contracts:
  - MockPolygonChild
  - LynxTACoChildApplication:
      proxy:
        contract_type: TACoChildApplication
      constructor:
        _rootApplication: $MockPolygonChild
        _minimumAuthorization: $FORTY_THOUSAND_TOKENS_IN_WEI_UNITS
  - LynxRitualToken:
      constructor:
        _totalSupplyOfTokens: $TEN_MILLION_TOKENS_IN_WEI_UNITS
        ...
 ```
 
 We could instead define:
 ```yml
 contracts:
  - MockPolygonChild
  - LynxTACoChildApplication:
      $proxy:
        $contract_type: TACoChildApplication
      _rootApplication: $MockPolygonChild
      _minimumAuthorization: $FORTY_THOUSAND_TOKENS_IN_WEI_UNITS
  - LynxRitualToken:
      _totalSupplyOfTokens: $TEN_MILLION_TOKENS_IN_WEI_UNITS
      ...
```

On one hand:
- much less expressive and much simpler file without needing "constructor:" dictionaries all the time when proxy is not even defined

on the other hand:
- co-mingling of proxy information with constructor parameters - slightly awkward, and perhaps slightly different from the "spirit" of yaml

Fair point/suggestion to raise by @cygnusv , and something we can discuss in this PR. Please provide any opinions you may have.

vzotova (approved):
LGTM
Tested with #167

KPrasch (approved):

KPrasch (approved):
Great documentation efforts here, too.  It will be useful to pull the PR description into a markdown document 🤠

deployment/params.py
```diff
@@ -27,183 +19,257 @@
...
+        self.contract_name = contract_name
+        self.check_for_proxy_instances = context.check_for_proxy_instances
+
+    def resolve(self) -> Any:
+        """Resolves a contract address."""
+        contract_container = get_contract_container(self.contract_name)
+        contract_instance = _get_contract_instance(contract_container)
+        if contract_instance == ZERO_ADDRESS:
+            # eager validation
+            return ZERO_ADDRESS
```
piotr-roslaniec, line 169: Should we also return `"0xdeadbeef"` here?

deployment/params.py
```diff
@@ -248,34 +309,184 @@ class ConstructorParameters:
...
+def _validate_transaction_args(
+    method: ContractTransactionHandler, args: typing.Tuple[Any, ...]
+) -> typing.Dict[str, Any]:
+    """Validates the transaction arguments against the function ABI."""
+    expected_length_abis = [abi for abi in method.abis if len(abi.inputs) == len(args)]
+    for abi in expected_length_abis:
+        named_args = {}
+        for arg, abi_input in zip(args, abi.inputs):
+            if not w3.is_encodable(abi_input.type, arg):
+                break
```
piotr-roslaniec, line 483: Perhaps we should throw here instead? Or is there a scenario where we knowingly take in a non-encodable param and we just want to skip it during this check?

cygnusv (changes requested):
First pass

deployment/params.py
derekpierre: This function is expected to return a contract address (`address` type) so `0xdeadbeef` would fail validation. The ZERO_ADDRESS is basically a placeholder for validating referenced contract address (contract variable) during initial eager validation before contracts are actually deployed.

deployment/params.py
derekpierre: Good question. 

Solidity supports function overloading, so there can be multiple methods with the same name, but the abi is different in terms of parameter length or parameter types. We initially filter out the different length cases, but then need to iterate through all resulting abis checking types. The first type mismatch we get, we don't want to raise because subsequent checks for other versions of the method may work.

Only if we make it out of the outer loop without ever returning the `else` block of the inner loop, then we know that there are no matching method abis, and can raise the exception.

deployment/params.py
```diff
@@ -27,183 +19,257 @@
...
+        deployer_account = Deployer.get_account()
+        if deployer_account is None:
+            return ZERO_ADDRESS
+        return deployer_account.address
 
-def _is_constant(variable: str) -> bool:
-    """Returns True if the variable is a deployment constant."""
-    return variable.isupper()
 
+class Constant(Variable):  # oxymoron
```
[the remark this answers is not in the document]
derekpierre: 😂

deployment/params.py
```diff
@@ -27,183 +19,257 @@
...
+    @staticmethod
+    def _get_call_data(variable, context) -> typing.Tuple[str, List[str], List[Any]]:
+        variable_elements = variable.split(",")
+        method_name = variable_elements[0]
+        method_args = list()
+        if len(variable_elements) > 1:
+            args = variable_elements[1:]
+            for arg in args:
+                processed_value = _process_raw_value(arg, context)
+                method_args.append(processed_value)
```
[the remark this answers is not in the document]
derekpierre: ✔️

deployment/params.py
```diff
@@ -27,183 +19,257 @@
...
+            if abi.name == method_name:
+                method_abi = abi
+        if not method_abi:
+            raise ValueError(f"ABI could not be found for method {contract_name}.{method_name}")
+        input_abi_types = [t.type for t in method_abi.inputs]
+        if len(input_abi_types) != len(method_args):
+            raise ValueError(
+                f"{contract_name}.{method_name} parameters length mismatch - "
+                f"ABI requires {len(input_abi_types)}, Got {len(method_args)}."
+            )
```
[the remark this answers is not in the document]
derekpierre: ✔️

cygnusv (approved):

manumonti (approved):
LGTM! ✌️

PR #171, opened 2023-11-01
Adds method to sample active staking providers using TACo child app

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Adds `getActiveStakingProviders` to `TACoChildApplication`

**Why it's needed:**
This method works using operator confirmations, and confirmations use bridge from polygon to mainnet. This direction needs bot or changes in dashboard. To decrease priority for that direction we can use child app for sampling

### DIFF
--- a/contracts/contracts/coordination/TACoChildApplication.sol
+++ b/contracts/contracts/coordination/TACoChildApplication.sol
@@ -143,36 +143,35 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
      * @param _maxStakingProviders Max providers for looking, if set 0 then all will be used
      * @return allAuthorizedTokens Sum of authorized tokens for active providers
      * @return activeStakingProviders Array of providers and their authorized tokens.
-     * Providers addresses stored as uint256
-     * @dev Note that activeStakingProviders[0] is an array of uint256, but you want addresses.
+     * Providers addresses stored together with amounts as bytes32
+     * @dev Note that activeStakingProviders is an array of bytes32, but you want addresses and amounts
      * Careful when used directly!
      */
     function getActiveStakingProviders(
         uint256 _startIndex,
         uint256 _maxStakingProviders
-    )
-        external
-        view
-        returns (uint256 allAuthorizedTokens, uint256[2][] memory activeStakingProviders)
-    {
+    ) external view returns (uint96 allAuthorizedTokens, bytes32[] memory activeStakingProviders) {
         uint256 endIndex = stakingProviders.length;
         require(_startIndex < endIndex, "Wrong start index");
         if (_maxStakingProviders != 0 && _startIndex + _maxStakingProviders < endIndex) {
             endIndex = _startIndex + _maxStakingProviders;
         }
-        activeStakingProviders = new uint256[2][](endIndex - _startIndex);
+        activeStakingProviders = new bytes32[](endIndex - _startIndex);
         allAuthorizedTokens = 0;
 
         uint256 resultIndex = 0;
         for (uint256 i = _startIndex; i < endIndex; i++) {
             address stakingProvider = stakingProviders[i];
             StakingProviderInfo storage info = stakingProviderInfo[stakingProvider];
-            uint256 eligibleAmount = info.authorized;
+            uint96 eligibleAmount = info.authorized;
             if (eligibleAmount < minimumAuthorization || !info.operatorConfirmed) {
                 continue;
             }
-            activeStakingProviders[resultIndex][0] = uint256(uint160(stakingProvider));
-            activeStakingProviders[resultIndex++][1] = eligibleAmount;
+            // bytes20 -> bytes32 adds padding after address: <address><12 zeros>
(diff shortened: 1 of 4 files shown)


contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -123,6 +130,56 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
...
+     * @notice Get the value of authorized tokens for active providers as well as providers and their authorized tokens
+     * @param _startIndex Start index for looking in providers array
+     * @param _maxStakingProviders Max providers for looking, if set 0 then all will be used
+     * @return allAuthorizedTokens Sum of authorized tokens for active providers
+     * @return activeStakingProviders Array of providers and their authorized tokens.
+     * Providers addresses stored as uint256
+     * @dev Note that activeStakingProviders[0] is an array of uint256, but you want addresses.
+     * Careful when used directly!
+     */
+    function getActiveStakingProviders(
```
derekpierre, line 151: In `nucypher-porter` and `nucypher` we currently do sampling using the same function on `TACoApplication` contract instead of `TACoChildApplication`. With this change, instead, we can sample using `TACoChildApplication`? Is there any difference between sampling using root vs child, other than using different chains? Is one potentially more up-to-date than the other?

I'm asking because, at the moment, Porter doesn't use/need any polygon endpoint, it only needs an eth endpoint. If using `TACoChildApplication` is better for some reason, then Porter will need to be updated to use polygon.

vzotova: yep, that's alternative to method in TACo root app until bot to sync from polygon to mainnet will be ready. 
Thing is - sampling returns only confirmed providers/operators. And without bot - all providers in root app will be unconfirmed. And considering that child has all data already - that's easiest option to start with

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -123,6 +130,56 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
...
+
+        uint256 resultIndex = 0;
+        for (uint256 i = _startIndex; i < endIndex; i++) {
+            address stakingProvider = stakingProviders[i];
+            StakingProviderInfo storage info = stakingProviderInfo[stakingProvider];
+            uint256 eligibleAmount = info.authorized;
+            if (eligibleAmount < minimumAuthorization || !info.operatorConfirmed) {
+                continue;
+            }
+            activeStakingProviders[resultIndex][0] = uint256(uint160(stakingProvider));
```
manumonti, line 175: Just out of curiosity: why returning `uint256` instead of `address`?

manumonti (approved):
LGTM ✌️! I've just left a question.

contracts/contracts/coordination/TACoChildApplication.sol
cygnusv: Because it's an array with 2 columns, one for address and the other for amount, so `uint256` seems like the more "common" type. 

@vzotova knowing that the address plus the amount (uint96) fit exactly in 1 word, how about returning a single uint256 value? This has the advantage that it would reduce the likelihood of using pagination by 50%, at the expense of requiring a very minor processing on the client side.

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -123,6 +130,56 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
...
+            activeStakingProviders[resultIndex++][1] = eligibleAmount;
+            allAuthorizedTokens += eligibleAmount;
+        }
+        assembly {
+            mstore(activeStakingProviders, resultIndex)
```
cygnusv, line 180: Why is this necessary?

vzotova: to cut empty spaces in returning array, otherwise it will always be `endIndex - _startIndex` with zeros in the end

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -19,6 +19,7 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
         bool operatorConfirmed;
         uint96 authorized;
         // TODO: what about undelegations etc?
+        uint256 index; // index in stakingProviders array + 1
```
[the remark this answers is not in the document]
vzotova: I tried 2 slots and it gives almost no gas savings (< 100 gas in average tx). Because it mostly about how and when we access those slots and less slots not always better. But I don't mind to make those changes
✔️

contracts/contracts/coordination/TACoChildApplication.sol
vzotova: ✔️

cygnusv (approved):

PR #175, opened 2023-11-06
Deploy TACo to dashboard domain and bump version

theref:


KPrasch (commented):
Pardon my ignorance, but why include the ABI at all if you are going to consume it from the @nucypher/nucypher-contracts NPM package?

theref:
> Pardon my ignorance, but why include the ABI at all if you are going to consume it from the @nucypher/nucypher-contracts NPM package?

Not sure I understand the question...in order for the abi to be on NPM it needs to be included in this repo right?

KPrasch:
> > Pardon my ignorance, but why include the ABI at all if you are going to consume it from the @nucypher/nucypher-contracts NPM package?
> 
> Not sure I understand the question...in order for the abi to be on NPM it needs to be included in this repo right?

😆 wow - I misread the repo name! Everything makes sense now.

PR #177, opened 2023-11-06
Ensure that TACoApplication is initialized for dashboard deployment

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

The TACoApplication contract needs to be initialized via TransparentUpgradeableProxy initialization.

Related to #170 .

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/deployment/constructor_params/dashboard/root.yml
+++ b/deployment/constructor_params/dashboard/root.yml
@@ -17,6 +17,8 @@ constants:
 contracts:
   - TACoApplication:
       proxy:
+        constructor:
+          _data: $encode:initialize
       constructor:
         _token: "0x3f16380656cAE45D3f80D8833682d2b606eD094A"
         _tStaking: "0x1da5d88C26EA4f87b5e09C3452eE2384Ee20DC75"


deployment/constructor_params/dashboard/root.yml
```diff
@@ -17,6 +17,8 @@ constants:
 contracts:
   - TACoApplication:
       proxy:
+        constructor:
+          _data: $encode:initialize
```
theref, line 21: what does `$encode` do?

derekpierre: It encodes the initialization call data for the proxy to execute. See #170 description, section called "Overriding _data to do an initialization call on underlying logic contract".

Basically, the Proxy on creation will call the `initialize` function on `TACoApplication` which takes no parameters. Previously, this was done in the deployment script as a separate step from proxy deployment, but now we can simply do it during proxy deployment.

theref (approved):

PR #179, opened 2023-11-09
TACoApplication: adds method `remainingAuthorizationDecreaseDelay`

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Upgrades solidity (bug fix), and adds method for dashboard

### DIFF
--- a/tests/application/test_authorization.py
+++ b/tests/application/test_authorization.py
@@ -329,6 +329,8 @@ def test_authorization_decrease_request(
     with ape.reverts():
         threshold_staking.authorizationDecreaseRequested(staking_provider, value, 1, sender=creator)
 
+    assert taco_application.remainingAuthorizationDecreaseDelay(staking_provider) == 0
+
     # Request of partial decrease
     tx = threshold_staking.authorizationDecreaseRequested(
         staking_provider, value, minimum_authorization, sender=creator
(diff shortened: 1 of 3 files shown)


theref (approved):

tests/application/test_authorization.py
```diff
@@ -450,8 +460,16 @@ def test_finish_authorization_decrease(
     with ape.reverts():
         taco_application.finishAuthorizationDecrease(staking_provider, sender=creator)
 
+    chain.pending_timestamp += deauthorization_duration // 4
+    remaining_duration = deauthorization_duration - deauthorization_duration // 4
+    assert (
+        taco_application.remainingAuthorizationDecreaseDelay(staking_provider)
+        == remaining_duration - 1
```
derekpierre, line 467: Is it `-1` because `current` is 1 less than `pending`?

derekpierre (approved):
🎸

tests/application/test_authorization.py
vzotova: it should opposite, current timestamp - next block after pending. To be honest that part is 100% on test chain and sometimes hard to predict in advance do we need to subtract second or not. At least it's stable

PR #191, opened 2023-11-16
Revamped lynx contract deployment

derekpierre:
Based over #189 
=============

**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Updated contracts on `lynx` to based on changes up to https://github.com/nucypher/nucypher-contracts/commit/820b8a40769a299db4f06462689de5cf0214f3bd.

Steps (custom changes, not to be committed, are catalogued here - https://github.com/derekpierre/nucypher-contracts/commits/david-derek-lynx-update):
- Modified lynx deployment scripts (root and child) to deploy updated MockPolygonRoot, TACoApplication (proxied), TACoChildApplication (proxied), Coordinator (proxied), GlobalAllowList, but reuse existing Lynx[Ritual/Staking]Token, TestnetThresholdStaking, MockPolygonChild contracts. See https://github.com/derekpierre/nucypher-contracts/commit/0234cdf79b1bc1f5a085a46efc1671945684d46c
- Merged resulting contract registry with existing lynx.json contract registry using the command:
```
$ ape run merge_registries --registry-1 ./deployment/artifacts/lynx.json --registry-2 ./deployment/artifacts/updated-lynx.json --output-registry ./deployment/artifacts/merged-lynx.json --deprecated-contract ProxyAdmin --deprecated-contract LynxTACoChildApplication --deprecated-contract TransparentUpgradeableProxy
```
- Modified `configure_staking` script and ran it to only perform `authorizationIncreased` and `bondOperator` on Goerli, and `updateAuthorization` and `updateOperator` on Mumbai, since application contracts were new - see https://github.com/derekpierre/nucypher-contracts/commit/8c1fc0e87ad3b3fec0457147f57781f9d09fe170
- Ran `confirm_operator_addresses` script as is.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

On a related note - `nucpyher/nucypher` will need to include the updated lynx.json contract registry before updating lynx nodes - related PR is TBD (cc @cygnusv )

### DIFF
--- a/deployment/artifacts/lynx.json
+++ b/deployment/artifacts/lynx.json
@@ -1,5956 +1,5259 @@
 {
-    "80001": {
-        "OpenAccessAuthorizer": {
-            "address": "0x23177458642a0Cf50000d8EAE53B0248475D3EdE",
-            "abi": [
-                {
-                    "type": "function",
-                    "name": "isAuthorized",
-                    "stateMutability": "pure",
-                    "inputs": [
-                        {
-                            "name": "",
-                            "type": "uint32",
-                            "internalType": "uint32"
-                        },
-                        {
-                            "name": "",
-                            "type": "bytes",
-                            "internalType": "bytes"
-                        },
-                        {
-                            "name": "",
-                            "type": "bytes",
-                            "internalType": "bytes"
-                        }
-                    ],
-                    "outputs": [
-                        {
-                            "name": "",
-                            "type": "bool",
-                            "internalType": "bool"
-                        }
-                    ]
-                }
-            ],
-            "tx_hash": "0x9ca790512e619a8cb9729c9be3a3e9a4f8d35b9b4a107aceb34ec3fc47b1a106",
-            "block_number": 41671721,
-            "deployer": "0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600"
+  "80001": {
+    "MockPolygonChild": {
+      "address": "0x45e06A2EaC4D928C8773A64b9eFe9d757B17F04D",
+      "abi": [
+        {
+          "type": "event",
+          "name": "AuthorizationUpdated",
+          "inputs": [
+            {
+              "name": "stakingProvider",
+              "type": "address",
+              "internalType": "address",
+              "indexed": true
+            },
+            {
+              "name": "amount",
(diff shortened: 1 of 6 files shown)


vzotova (approved):

deployment/artifacts/lynx.json
```diff
@@ -1462,12 +1639,12 @@
                     "outputs": []
                 }
             ],
-            "tx_hash": "0x6495205ffa4f622cef2824c31b615812934379db8a6766e345881f5ac635f237",
-            "block_number": "40514266",
+            "tx_hash": "0xc2fd6f00ad2881cf3fd6b3440746dda9811c2cde625ef6c6d9acf47886bf5b77",
+            "block_number": 42479020,
```
cygnusv, line 1643: Block numbers seem to be now int literals, rather than strings, which is fine in itself, but I'm wondering about potential issues when parsing the registry.

cygnusv (commented):
Looks good, just a minor comment regarding format of registry.

deployment/artifacts/lynx.json
derekpierre: That's strange that it was a string before. I think that might be a bug with retroactively obtaining the deployment information - that was functionality that was added after a deployment was done (and a registry without that information already generated) via the `get_creation_info` - https://github.com/nucypher/nucypher-contracts/blob/main/deployment/legacy.py#L43. I think that line needs to cast the value to an int. 
7c515233f4d. Filed #194 - it may/may not apply to `mainnet` deployment if we merge the new deployment registry with the `mainnet` legacy registry (not sure if applicable or not).

In any case, I think we are good:
- On the `nucypher-contracts`-side, the `RegistryEntry` object expects `block_number` to be an `int` - https://github.com/nucypher/nucypher-contracts/blob/main/deployment/registry.py#L25, so writing it out it should be an int.
- On the `nucypher`-side, it doesn't care about block_number - https://github.com/nucypher/nucypher/blob/v7.0.0/nucypher/blockchain/eth/registry.py#L331

Issue #194, opened 2023-11-17
Obtaining creation information using legacy.py should cast block_number to an `int`


derekpierre:
Potential bug noticed here - https://github.com/nucypher/nucypher-contracts/pull/191#discussion_r1397140147.

Related to #135 .

derekpierre:
Fixed via #196.

PR #196, opened 2023-11-19
Normalize deployment registries

### LINKED ISSUE #192: Introduce canonical format for registries
* Order: registry entities should follow a common order in all the hierarchy levels. I'd recommend just doing it alphabetically, except for the contract dictionary, where it's convenient that `address` is the first entry. For the ABI part, which on JSON is a list of dictionaries, I'd order it by first the value of the type `entry` on each dictionary, second by the `name` entry.
* Identation: just pick something and be consistent with that (e.g., 2 spaces)
* ABI format: This is where it can be more tricky. Some tools will give you more verbose ABIs than others (e.g., `internalType`). Needs investigation.

Why doing this? Mostly, for reviewing changes on the registry

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
* Adds script to normalize registries according #192
* Uses it to normalize existing registries.
* Removes `open_access_authorizer.json` registry

In particular this enforces the following when writing registries:
* Chain ID entries (as strings, because ints can't be JSON keys) are lexicographically ordered (e.g., "1" before "11", but "11" before "5")
* Within each chain dictionary, contract entries are lexicographically ordered by contract name.
* Contract entries are dictionaries, withe following sequence of keys: `address`, `abi`, `tx_hash`, `block_number`, `deployer`.
* Each ABI is a list of dictionaries, lexicographically sorted by their `"type"` and `"name"` keys.
* JSON file uses 4 spaces for indentation, as well as `','` and `': '` as separators.
* Block numbers are expressed as ints, not strings

**Issues fixed/closed:**
- Closes #192 

**Why it's needed:**
This will facilitate reviewing changes to registry files

**Notes for reviewers:**


cygnusv:
Example of how `mainnet.json` will look like once we deploy child contracts, and merge with existing registry:
image
As you can see, resulting registry follows the expected order.

cygnusv:
> Looks great! What is the expected workflow when it can't automatically merge registries and instead creates a new `*.unmerged.json`?

Good question. Since most of these changes are directly enforced in the `write_registry()` function, running `ape run merge_registries` will take care of that too.

piotr-roslaniec (approved):

deployment/registry.py
```diff
@@ -97,20 +102,27 @@ def read_registry(filepath: Path) -> List[RegistryEntry]:
...
+        entry_abi = list(entry.abi)
+        entry_abi.sort(key=lambda d: (d["type"], d.get("name", "")))
+
         data[str(entry.chain_id)][entry.name] = {
             "address": entry.address,
-            "abi": entry.abi,
+            "abi": entry_abi,
             "tx_hash": entry.tx_hash,
-            "block_number": entry.block_number,
+            "block_number": int(entry.block_number),
```
derekpierre, line 125: 👍 - this indirectly takes care of https://github.com/nucypher/nucypher-contracts/issues/194.

scripts/normalize_registry.py
```diff
@@ -0,0 +1,17 @@
...
+@click.command()
+@click.option(
+    "--registry",
+    help="Filepath to registry file",
+    type=click.Path(dir_okay=False, exists=True, path_type=Path),
+    required=True,
+)
+def cli(registry):
+    """Normalize registry file"""
+    normalize_registry(registry)
```
derekpierre, line 17: 💥

derekpierre (approved):
🎸

PR #128, opened 2023-10-02
TACo Mainnet deployment parameters and intermediate artifacts

cygnusv:
(Depends on #196, wait for it to be merged)

Specially important for reviewers:
* FxPortal Deployment info for Mainnet: FxRoot, FxChild, CheckpointManager (See https://wiki.polygon.technology/docs/pos/design/bridge/l1-l2-communication/state-transfer/#prerequisites)
* DAI address on Polygon: https://polygonscan.com/token/0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063
* Initial fee rate, which is 35 cents per day but expressed in DAI units per second. In Python:
```py
>>> 35*10**16//86400
4050925925925
```
-----

**Update**: We used these parameters to deploy TACo contracts on Ethereum Mainnet and Polygon Mainnet. Intermediary artifacts are included in this PR, as we need additional verification before confirming this is the official deployment.

PR #200, opened 2023-11-21
TACo Mainnet Intermediate Deployment Artifacts

cygnusv:
Let's use this PR as a safe checkpoint after the deployment process. Now registries needs merging.


cygnusv:
Actually, let's just do this in #128

PR #207, opened 2023-11-27
StakingEscrow: removes unused code

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
Now when all legacy tokens unstaked we can remove unused parts of code which will allow to clean `TokenStaking` contract

### DIFF
--- a/contracts/contracts/StakingEscrow.sol
+++ b/contracts/contracts/StakingEscrow.sol
@@ -36,48 +36,6 @@ interface IVendingMachine {
 
 
 /**
-* @title StakingEscrowStub
-* @notice Stub is used to deploy main StakingEscrow after all other contract and make some variables immutable
-* @dev |v1.1.0|
-*/
-contract StakingEscrowStub is Upgradeable {
-    NuCypherToken public immutable token;
-    // only to deploy WorkLock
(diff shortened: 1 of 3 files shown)


contracts/contracts/StakingEscrow.sol
```diff
@@ -202,7 +158,7 @@ contract StakingEscrow is Upgradeable, IERC900History {
         IVendingMachine _vendingMachine
     ) {
         require(_token.totalSupply() > 0 &&
-            _tStaking.stakedNu(address(0)) == 0 &&
+            _tStaking.getApplicationsLength() != 0 &&
```
cygnusv, line 161: why not checking that `_tStaking.token() == address(_token)` instead? seems more natural at first

cygnusv (approved):

contracts/contracts/StakingEscrow.sol
vzotova: I wanted, but it's not part of `IStaking`

PR #209, opened 2023-11-28
Upgrade scripts

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


KPrasch (approved):

deployment/params.py
```diff
@@ -614,6 +615,31 @@ def _deploy_proxy(
...
+            raise ValueError(
+                f"Admin slot for contract at {proxy_address} is empty. "
+                "Are you sure this is an EIP1967-compatible proxy?"
+            )
+        
+        admin_address = to_checksum_address(admin_slot[-20:])
+        proxy_admin = OZ_DEPENDENCY.ProxyAdmin.at(admin_address)
+        # TODO: Check that owner of proxy admin is deployer
+
+        implementation = self.deploy(container)
```
derekpierre, line 635: @cygnusv Not a big deal given our current situation works this way, but previously the container name could be different from the contract name to deploy eg. `LynxTACoChildApplication` was the contract to deploy, but the container name that we wanted to use for it when proxied was `TACoChildApplication`.

Definitely doesn't affect us currently since we deploy the actual contracts and not modified versions of them, but just something of note in case that changes.

deployment/params.py
```diff
@@ -614,6 +615,31 @@ def _deploy_proxy(
         )
         return contract_type_container.at(proxy_contract.address)
 
+    def upgrade(self, container: ContractContainer, proxy_address, data=b'') -> ContractInstance:
```
derekpierre, line 618: Very cool!

PR #211, opened 2023-11-30
fix(test): fix ts registry tests not running all test cases

piotr-roslaniec:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -38,6 +38,10 @@ export const getContract = (
   chainId: ChainId | number,
   contract: ContractName | string,
 ): ChecksumAddress => {
+  if (!contractNames.includes(contract as ContractName)) {
+    throw new Error(`Invalid contract name: ${contract}`);
+  }
+
   const registry = domainRegistry[domain];
   if (!registry) {
     throw new Error(`No contract registry found for domain: ${domain}`);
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -1,47 +1,47 @@
...
+export type ContractRegistry = Record<string, Record<string, DeployedContract>>;
 
 export const domainRegistry: Record<string, ContractRegistry> = {
   lynx: lynxRegistryJson,
   tapir: tapirRegistryJson,
   mainnet: mainnetRegistryJson,
 };
 
-export type Domain = 'mainnet' | 'oryx' | 'tapir' | 'lynx'
+export type Domain = "mainnet" | "oryx" | "tapir" | "lynx";
```
derekpierre, line 30: @piotr-roslaniec Should we get rid of `oryx` here, or is that still needed for something?

piotr-roslaniec: Hmm, it's not being used anymore for PRE right?

derekpierre: Only Masterfile uses oryx (although I'm not sure how much), but I thought our latest code already deprecated `oryx` anyway i.e. Masterfile must use an old version of `nucypher-ts`/`taco-web` with `oryx`.

(cc @cygnusv , @KPrasch  in case I'm mistaken)

PR #214, opened 2023-12-04
Return minimumAuthorization instead of _minimumAuthorization

### LINKED ISSUE #213: Return `minimumAuthorization` instead of `_minimumAuthorization`
https://github.com/nucypher/nucypher-contracts/blob/5815446b7b8c307ab2bd5b066a9192815648653b/contracts/contracts/TACoApplication.sol#L323C42-L323C42

theref:
Fixes #213

**Type of PR:**
- [x] Bugfix

**Required reviews:** 
- [x] 1

variables prefixed with underscores are for internal use, an `external` `view` function should return variables without the underscore prefix

### DIFF
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -300,7 +300,7 @@ contract TACoApplication is
      *  @notice Returns authorization-related parameters of the application.
      *  @dev The minimum authorization is also returned by `minimumAuthorization()`
      *       function, as a requirement of `IApplication` interface.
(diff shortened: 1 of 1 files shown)


vzotova (changes requested):
`minimumAuthorization` shadows main getter with the same name. Even it's warning - I strongly agree with it. I can't say I have easy solution but we should discuss it. And especially in https://github.com/threshold-network/solidity-contracts/pull/157

theref:
ok, I'm happy to put this on the back burner for now because I can work around it in the dashboard

PR #188, opened 2023-11-15
Beta Program Initiator

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
> High-level idea of the changes introduced in this PR. 
> List relevant API changes (if any), as well as related PRs and issues.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #222, opened 2024-01-10
`BetaProgramInitiator` deployment

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

Related to #188 

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/deployment/constructor_params/lynx/beta_program_initiator.yml
+++ b/deployment/constructor_params/lynx/beta_program_initiator.yml
@@ -17,4 +17,4 @@ contracts:
     - BetaProgramInitiator:
         constructor:
           _coordinator: $COORDINATOR_PROXY
-          _executor: EXECUTOR
+          _executor: $EXECUTOR
--- a/deployment/constructor_params/lynx/beta_program_initiator.yml
+++ b/deployment/constructor_params/lynx/beta_program_initiator.yml
@@ -1,0 +1,20 @@
+deployment:
+  name: beta-program-initiator
+  chain_id: 80001
+
+artifacts:
+  dir: ./deployment/artifacts/
+  filename: beta_program_initiator_lynx.json
+
+constants:
+  # See deployment/artifacts/lynx.json
+  COORDINATOR_PROXY: "0x530608219a8A671FD183534b17E2a2CE09e782a4"
+
+  # lynx deployer account
+  EXECUTOR: "0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600"
+
+contracts:
+    - BetaProgramInitiator:
+        constructor:
+          _coordinator: $COORDINATOR_PROXY
+          _executor: EXECUTOR
--- a/deployment/constructor_params/tapir/beta_program_initiator.yml
+++ b/deployment/constructor_params/tapir/beta_program_initiator.yml
@@ -17,4 +17,4 @@ contracts:
     - BetaProgramInitiator:
         constructor:
           _coordinator: $COORDINATOR_PROXY
-          _executor: EXECUTOR
+          _executor: $EXECUTOR
--- a/deployment/constructor_params/tapir/beta_program_initiator.yml
+++ b/deployment/constructor_params/tapir/beta_program_initiator.yml
@@ -1,0 +1,20 @@
+deployment:
+  name: beta-program-initiator
+  chain_id: 80001
+
+artifacts:
+  dir: ./deployment/artifacts/
+  filename: beta_program_initiator_tapir.json
+
+constants:
+  # See deployment/artifacts/tapir.json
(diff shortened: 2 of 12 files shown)


deployment/constructor_params/lynx/beta_program_initiator.yml
```diff
@@ -0,0 +1,20 @@
...
+artifacts:
+  dir: ./deployment/artifacts/
+  filename: beta_program_initiator_lynx.json
+
+constants:
+  # See deployment/artifacts/lynx.json
+  COORDINATOR_PROXY: "0x530608219a8A671FD183534b17E2a2CE09e782a4"
+
+  # lynx deployer account
+  EXECUTOR: "0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600"
```
derekpierre, line 14: @cygnusv , @KPrasch wdyt about using the deployer as the executor `lynx`? Open to other ideas.

deployment/constructor_params/tapir/beta_program_initiator.yml
```diff
@@ -0,0 +1,20 @@
...
+artifacts:
+  dir: ./deployment/artifacts/
+  filename: beta_program_initiator_tapir.json
+
+constants:
+  # See deployment/artifacts/tapir.json
+  COORDINATOR_PROXY: "0xdED71c37e4e17aF2c825c2A4441Dd6BF5A98D194"
+
+  # tapir deployer account
+  EXECUTOR: "0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600"
```
derekpierre, line 14: @cygnusv , @KPrasch wdyt about using the deployer as the executor on `tapir`? Open to other ideas.

deployment/constructor_params/lynx/beta_program_initiator.yml
cygnusv: Yes, that's the most sensible choice for this, as well as Tapir

deployment/constructor_params/tapir/beta_program_initiator.yml
cygnusv: 👍

derekpierre:
Lynx:
- BetaProgramInitiator (Verified): https://mumbai.polygonscan.com/address/0xb6e6512F4c9A74Ef1b742d832852769F3CEE646c#code
- Granted initiator role: 
https://mumbai.polygonscan.com/tx/0x67cf6b0b4dfaa1e2634cf97259e9256bb9261a8fca84b5b9640d305e81c882fa

**NOTE:** We may need to update the Coordinator implementation contract on `Lynx` - it didn't have the `feeDeduction` method needed by the `BetaProgramInitiator` contract.

Tapir:
- BetaProgramInitiator (Verified):
https://mumbai.polygonscan.com/address/0x070fbc5a82776A1fd9945824d0719E4E7a5Ac01c#code
- Granted initiator role: https://mumbai.polygonscan.com/tx/0x352450f0f3c4bf0f08757b0eb94b3420103f2a76941e8b52861d6e097dcc2c6f

derekpierre:
Fresh `Coordinator` implementation contract deployed at `0x90E112651A57F1484279b9da5C0Bd3380E59ceEA` for `lynx`, and `lynx` `Coordinator` proxy contract updated to point to new contract.

vzotova (approved):

derekpierre:
> We need to include these changes in this PR, right?

Yes. One minor update to the ABI in the registry for the `feeDeduction` method - included now.

> We don't need to coordinate it with nucypher/nucypher, right?

From the `nucypher/nucypher` side it's not needed immediately for functionality because the `feeDeduction` method (the reason for the update) is not used by the nodes/python code. It is used by the `BetaProgramInitiator` contract which we can now test on `lynx` and `tapir`.

> And that would mandate a bump in @nucypher/nucypher-contracts, na hence @nucypher/taco?

I suspect `taco-web` doesn't use `feeDeduction` either - but you can let me know if it does. 

An update will occur eventually as part of 7.1.x, but I don't think it needs to be immediate because neither client nor server-side need it at the moment.

piotr-roslaniec:
> I suspect taco-web doesn't use feeDeduction either - but you can let me know if it does.

It doesn't, but I still need to use the new `Coordinator` contract for everything else.

Ok, I think I can implement it now without breaking anything / waiting for 7.1.x.

derekpierre:
> It doesn't, but I still need to use the new Coordinator contract for everything else.

Coordinator is proxied so you are still using the same proxy contract, it's the underlying implementation contract that changed.

cygnusv (commented):
Can't approve because I opened it, but @derekpierre's commits look good to me.

PR #231, opened 2024-01-26
Optimized Participant Views

KPrasch:
- Paginated `getParticipants`view with optional transcripts
- `isParticipant` checks if a staking provider address belong to a ritual's participants
- Optimize internal participant lookups with binary search

##### Notes
Requires downstream changes https://github.com/nucypher/nucypher/pull/3419

### DIFF
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -104,6 +104,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     IReimbursementPool internal reimbursementPool;
     mapping(address => ParticipantKey[]) internal participantKeysHistory;
     mapping(bytes32 => uint32) internal ritualPublicKeyRegistry;
+    Participant internal SENTINEL_PARTICIPANT;
 
     constructor(
         ITACoChildApplication _application,
@@ -422,8 +423,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     function getParticipantIndex(
         Ritual storage ritual,
         address provider
-    ) internal view returns (int256) {
-        if (ritual.participant.length == 0) {return -1;}
+    ) internal view returns (bool, uint256, Participant storage participant) {
         uint256 low = 0;
         uint256 high = ritual.participant.length - 1;
         while (low <= high) {
@@ -432,36 +432,50 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
             if (participant.provider < provider) {
                 low = mid + 1;
             } else if (participant.provider > provider) {
-                if (mid == 0) {return -1;}
                 high = mid - 1;
             } else {
-                return int256(mid);
+                return (true, mid, participant);
             }
         }
-        return -1;
+        return (false, 0, SENTINEL_PARTICIPANT);
     }
 
-    function getParticipantFromProvider(uint32 ritualId, address provider) external view returns (Participant memory participant) {
+    function getParticipantIndex (
+        uint32 ritualId,
+        address provider
+    ) external view returns (uint256) {
         Ritual storage ritual = rituals[ritualId];
-        int256 index = getParticipantIndex(ritual, provider);
-        require(index >= 0, "Participant not found");
-        return ritual.participant[uint256(index)];
+        (bool found, uint256 index,) = getParticipantIndex(ritual, provider);
+        require(found, "Participant not found");
+        return index;
+    }
+
+    function isParticipant(uint32 ritualId, address provider) external view returns (bool) {
+        Ritual storage ritual = rituals[ritualId];
+        (bool found,,) = getParticipantIndex(ritual, provider);
+        return found;
+    }
+
+    function getParticipantFromProvider(
+        uint32 ritualId,
+        address provider
+    ) external view returns (Participant memory) {
+        Ritual storage ritual = rituals[ritualId];
+        (bool found,, Participant storage participant) = getParticipantIndex(ritual, provider);
+        require(found, "Participant not found");
+        return participant;
     }
 
     function getParticipantFromProvider(
         Ritual storage ritual,
         address provider
     ) internal view returns (Participant storage participant) {
-        int256 index = getParticipantIndex(ritual, provider);
-        require(index >= 0, "Participant not found");
-        return ritual.participant[uint256(index)];
+        (bool found, uint256 index, Participant storage participant) = getParticipantIndex(ritual, provider);
+        require(found, "Participant not found");
+        return participant;
     }
 
-    function isParticipant(uint32 ritualId, address provider) external view returns (bool) {
-        Ritual storage ritual = rituals[ritualId];
-        int256 index = getParticipantIndex(ritual, provider);
-        return index >= 0;
-    }
+
 
     function getParticipants(
         uint32 ritualId
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -311,7 +311,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         );
 
         address provider = application.operatorToStakingProvider(msg.sender);
-        Participant storage participant = getParticipantFromProvider(ritual, provider);
+        Participant storage participant = getParticipant(ritual, provider);
 
         require(application.authorizedStake(provider) > 0, "Not enough authorization");
         require(participant.transcript.length == 0, "Node already posted transcript");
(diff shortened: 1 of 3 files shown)


contracts/contracts/coordination/Coordinator.sol
```diff
@@ -424,26 +425,105 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+    }
+
+    function getParticipant(
+        Ritual storage ritual,
+        address provider
+    ) internal view returns (Participant storage) {
+        (bool found, , Participant storage participant) = findParticipant(ritual, provider);
+        if (!found) {
+            revert("Participant not part of ritual");
+        }
```
vzotova, line 462:
```suggestion
        require(found, "Participant not part of ritual");
```

vzotova:
general question regarding those optimizations: would this add at any moment call for data in a loop? Why I'm asking - similar task was with `getActiveStakikingProviders` and in the result best outcome was with less calls as possible

KPrasch (commented):
There is still a failing test on CI - but I've marked this PR as RFR to get the review process started ahead of finalization these changes to invite feedback.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -104,6 +104,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     IReimbursementPool internal reimbursementPool;
     mapping(address => ParticipantKey[]) internal participantKeysHistory;
     mapping(bytes32 => uint32) internal ritualPublicKeyRegistry;
+    Participant internal sentinelParticipant;
```
cygnusv, line 107: In order to avoid potential problems later with storage layout if `Participant` struct changes and/or more state variables are added to the contract, we can change the current sentinel for the following:
```suggestion
    mapping(uint256 => Participant) internal sentinelParticipants;
    uint256 internal sentinelIndex = 0;
```
Later on, we just need to assign `sentinelParticipant = sentinelParticipants[sentinelIndex] when needed`.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -424,26 +425,101 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+        (bool found, , Participant storage participant) = findParticipant(ritual, provider);
+        require(found, "Participant not part of ritual");
+        return participant;
+    }
+
+    function getParticipant(
+        uint32 ritualId,
+        address provider,
+        bool transcript
+    ) public view returns (Participant memory, uint256) {
```
cygnusv, line 468: this function doesn't seem to be used internally, right? In that case, it could be set to external.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -424,26 +425,101 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
-        for (uint256 i = 0; i < length; i++) {
-            Participant storage participant = ritual.participant[i];
+        if (length == 0) {
+            return (false, 0, sentinelParticipant);
+        }
+        uint256 low = 0;
+        uint256 high = length - 1;
+        while (low <= high) {
+            uint256 mid = (low + high) / 2;
+            Participant storage participant = ritual.participant[mid];
```
cygnusv, line 440: I think this makes things clearer but you can take it or leave it:
```suggestion
            Participant storage middleParticipant = ritual.participant[mid];
```

contracts/contracts/coordination/Coordinator.sol
derekpierre: ✔️

contracts/contracts/coordination/Coordinator.sol
derekpierre: Updated - let us know if we did it correct (cc @vzotova ).

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -424,26 +425,103 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+        if (!transcript) {
+            participant.transcript = "";
+        }
+        return (participant, index);
     }
 
-    function getParticipantFromProvider(
+    function getParticipants(
+        uint32 ritualId,
+        uint256 _startIndex,
```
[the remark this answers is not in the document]
derekpierre: 👍 Went the no underscores route since most of the existing functions didn't use it.

contracts/contracts/coordination/Coordinator.sol
vzotova: I do not see anything wrong, but also I don't get everything about those variables

derekpierre (approved):
Implicit approval since I'm a co-author of the PR.

contracts/contracts/coordination/Coordinator.sol
derekpierre: @cygnusv 's initial concern was around the usage of a `sentinelParticipant` slot for a Participant struct which could change size/structure later on down the road.

By using a mapping we don't have to worry about the struct/slot potentially changing size.

(@cygnusv hopefully I explained this properly 😅 )

vzotova: I got the idea, in that case do we need index field for that? if value always will be only one then it's not necessary
other way to do that, a bit more straightforward -> put some gap between regular `sentinelParticipant` struct and previous values, like `uint256[20]`, in that case extending will be same easy, and gap is not taking any gas

derekpierre: We (@KPrasch and myself) defer to you and @cygnusv for your expertise on the best course of action here ... happy to make adjustments as needed.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -424,26 +431,103 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+                ritualParticipant.transcript = "";
             }
+            ritualParticipants[resultIndex++] = ritualParticipant;
         }
-        revert("Participant not part of ritual");
+
+        return ritualParticipants;
     }
 
-    function getParticipantFromProvider(
```
derekpierre, line 442: This rename will cause 7.0.4 nodes to no longer work. If we want to maintain backwards compatibility we should keep the original name.

piotr-roslaniec:
Note: I don't see any breaking changes for `taco-web`

contracts/contracts/coordination/Coordinator.sol
derekpierre: ✔️

contracts/contracts/coordination/Coordinator.sol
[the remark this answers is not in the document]
derekpierre: ✔️

vzotova (approved):
LGTM
There is still one linting issue and one my suggestion, but that is minor

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -424,26 +425,103 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         return ritual.publicKey;
     }
 
-    function getParticipantFromProvider(
+    function findParticipant(
         Ritual storage ritual,
         address provider
-    ) internal view returns (Participant storage) {
+    ) internal view returns (bool, uint256, Participant storage participant) {
```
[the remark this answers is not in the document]
derekpierre: We're having some testing issues with ape and named variables as it pertains to how arrays/named tuples are processed by ape as output values - see https://github.com/KPrasch/nucypher-contracts/blob/main/tests/test_coordinator.py#L361. @cygnusv has been doing some investigation - may/may not be a bug with ape.

theref:
Could we clean up the commit history? i think a bunch of these can be squashed together - then it looks good to go

derekpierre:
> Could we clean up the commit history? i think a bunch of these can be squashed together - then it looks good to go

I cleaned up some of the extra linting commits - it's a little tricky doing more than that because the commits are very specific, and there was a bit of winding journey  (😄 ) that would get lost as part of a squashing which I don't think we want to lose because context would be lost.

cygnusv:
> > Could we clean up the commit history? i think a bunch of these can be squashed together - then it looks good to go
> 
> I cleaned up some of the extra linting commits - it's a little tricky doing more than that because the commits are very specific, and there was a bit of winding journey (😄 ) that would get lost as part of a squashing which I don't think we want to lose because context would be lost.

Yeah, the journey is important. Unless the commit history is particularly messy, I'd prefer not to squash it.

tests/test_coordinator.py
```diff
@@ -318,6 +319,110 @@ def test_post_transcript_but_not_waiting_for_transcripts(
...
+    assert len(participants) == len(nodes)
+    for index, participant in enumerate(participants):
+        assert participant.provider == nodes[index].address
+        assert participant.aggregated is False
+        assert participant.transcript == transcript
+
+    # n at a time
+    for n_at_a_time in range(2, len(nodes) // 2):
+        if len(nodes) % n_at_a_time == 1:
+            # TODO ugly; decoding issues with web3py/ape w.r.t
```
cygnusv, line 362: Remove workaround for tuple array decoding issue with ape

cygnusv (approved):

PR #240, opened 2024-02-15
Adds deauthorizing amount to bridge synchronization

vzotova:
**Type of PR:**
- [X] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Root -> Child TACo bridge now syncs deauth amount and end of deauth. 

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
To allow confirmation based on authorized amount only (without deauth amount)

**Notes for reviewers:**
TACo child upgrade should be done first, otherwise bridge syncing could be an issue. Even if it will occur it can be solved with calling manual method later

### DIFF
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -150,6 +150,8 @@ contract TACoApplication is
     event ManualChildSynchronizationSent(
         address indexed stakingProvider,
         uint96 authorized,
+        uint96 deauthorizing,
+        uint64 endDeauthorization,
         address operator
     );
 
@@ -655,10 +657,18 @@ contract TACoApplication is
     }
 
     /**
-     * @notice Get all tokens delegated to the staking provider
+     * @notice Get eligible stake delegated to the staking provider
      */
-    function getEligibleAmount(StakingProviderInfo storage _info) internal view returns (uint96) {
-        return _info.authorized - _info.deauthorizing;
+    function eligibleStake(
+        address _stakingProvider,
+        uint256 _endDate
+    ) public view returns (uint96) {
+        StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
+        uint96 eligibleAmount = info.endDeauthorization == 0 || info.endDeauthorization >= _endDate
+            ? info.authorized
+            : info.authorized - info.deauthorizing;
+
+        return eligibleAmount;
     }
 
     /**
@@ -690,6 +700,7 @@ contract TACoApplication is
      * @notice Get the value of authorized tokens for active providers as well as providers and their authorized tokens
      * @param _startIndex Start index for looking in providers array
      * @param _maxStakingProviders Max providers for looking, if set 0 then all will be used
+     * @param _cohortDuration Duration during which staking provider should be active. 0 means no end date
      * @return allAuthorizedTokens Sum of authorized tokens for active providers
      * @return activeStakingProviders Array of providers and their authorized tokens.
      * Providers addresses stored together with amounts as bytes32
@@ -698,7 +709,8 @@ contract TACoApplication is
      */
     function getActiveStakingProviders(
         uint256 _startIndex,
-        uint256 _maxStakingProviders
+        uint256 _maxStakingProviders,
+        uint32 _cohortDuration
     ) external view returns (uint256 allAuthorizedTokens, bytes32[] memory activeStakingProviders) {
         uint256 endIndex = stakingProviders.length;
         require(_startIndex < endIndex, "Wrong start index");
@@ -707,12 +719,15 @@ contract TACoApplication is
         }
         activeStakingProviders = new bytes32[](endIndex - _startIndex);
         allAuthorizedTokens = 0;
+        uint256 endDate = _cohortDuration == 0
+            ? type(uint256).max
+            : block.timestamp + _cohortDuration;
 
         uint256 resultIndex = 0;
         for (uint256 i = _startIndex; i < endIndex; i++) {
             address stakingProvider = stakingProviders[i];
             StakingProviderInfo storage info = stakingProviderInfo[stakingProvider];
-            uint256 eligibleAmount = getEligibleAmount(info);
+            uint256 eligibleAmount = eligibleStake(stakingProvider, endDate);
             if (eligibleAmount < minimumAuthorization || !info.operatorConfirmed) {
                 continue;
             }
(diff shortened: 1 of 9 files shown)


contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -141,6 +187,7 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
      * @notice Get the value of authorized tokens for active providers as well as providers and their authorized tokens
      * @param _startIndex Start index for looking in providers array
      * @param _maxStakingProviders Max providers for looking, if set 0 then all will be used
+     * @param _cohortDuration Duration during which staking provider should be active. 0 means no end date
```
manumonti, line 190: just asking: does this sound more crystal clear to you?
```suggestion
     * @param _cohortDuration Duration during which staking provider should be active. 0 means forever
```

contracts/contracts/TACoApplication.sol
```diff
@@ -655,10 +657,18 @@ contract TACoApplication is
...
-    function getEligibleAmount(StakingProviderInfo storage _info) internal view returns (uint96) {
-        return _info.authorized - _info.deauthorizing;
+    function eligibleStake(
+        address _stakingProvider,
+        uint256 _endDate
+    ) public view returns (uint96) {
+        StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
+        uint96 eligibleAmount = info.endDeauthorization == 0 || info.endDeauthorization >= _endDate
+            ? info.authorized
+            : info.authorized - info.deauthorizing;
```
cygnusv, line 669: Can we make this code a bit clearer?
```suggestion
        uint96 eligibleAmount = info.authorized;
        if (0 < info.endDeauthorization && info.endDeauthorization < _endDate) {
            eligibleAmount -= info.deauthorizing;
        }
```
Additionally, I'd introduce the restriction that `require(_endDate > block.timestamp);`

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -66,18 +68,49 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
...
+    }
+
+    function eligibleStake(
+        address _stakingProvider,
+        uint256 _endDate
+    ) public view returns (uint96) {
+        StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
+        uint96 eligibleAmount = info.endDeauthorization == 0 || info.endDeauthorization >= _endDate
+            ? info.authorized
+            : info.authorized - info.deauthorizing;
```
cygnusv, line 87: Same as above.

contracts/contracts/TACoApplication.sol
```diff
@@ -655,10 +657,18 @@ contract TACoApplication is
     }
 
     /**
-     * @notice Get all tokens delegated to the staking provider
+     * @notice Get eligible stake delegated to the staking provider
      */
```
manumonti, line 661:
```suggestion
    /**  
     * @notice Returns the amount of stake that are going to be effectively  
     *         staked until the specified date. I.e: in case a deauthorization  
     *         is going to be made during this period, the returned amount will 
     *         be the staked amount minus the deauthorizing amount.  
     */  
```

Same comment that in TACoChildApplication.sol:

contracts/contracts/TACoApplication.sol
```diff
@@ -690,6 +700,7 @@ contract TACoApplication is
      * @notice Get the value of authorized tokens for active providers as well as providers and their authorized tokens
      * @param _startIndex Start index for looking in providers array
      * @param _maxStakingProviders Max providers for looking, if set 0 then all will be used
+     * @param _cohortDuration Duration during which staking provider should be active. 0 means no end date
```
manumonti, line 703: Same comment in `TACoChildApplication.sol`
```suggestion
     * @param _cohortDuration Duration during which staking provider should be active. 0 means forever
```

manumonti (approved):
LGTM! I've just left some suggestions.

contracts/contracts/TACoApplication.sol
vzotova: ✔️

contracts/contracts/TACoApplication.sol
vzotova: ✔️

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -66,18 +68,49 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
...
+    /**
+     * @notice Returns the amount of stake that is pending authorization
+     *         decrease for the given staking provider. If no authorization
+     *         decrease has been requested, returns zero.
+     */
+    function pendingAuthorizationDecrease(address _stakingProvider) external view returns (uint96) {
+        return stakingProviderInfo[_stakingProvider].deauthorizing;
+    }
+
+    function eligibleStake(
```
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/coordination/TACoChildApplication.sol
vzotova: ✔️

contracts/contracts/TACoApplication.sol
vzotova: ✔️ 
i'm not sure about `require` to be honest, it's a call and even called with wrong data it won't break logic

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -178,6 +229,14 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
             mstore(activeStakingProviders, resultIndex)
         }
     }
+
+    // TODO only for backward compatibility
+    function getActiveStakingProviders(
```
derekpierre, line 242: While this maintains backwards compatibility for now - presumably we will need to updates to `nucypher/nucypher` and `nucypher/nucypher-porter` for proper sampling. 

Particularly in Porter's case, the `/bucket_sampling` and `/get_ursulas` (although not sure we still need this one) endpoint(s) will need to take a duration.

derekpierre (approved):
🎸

tests/application/test_authorization.py
```diff
@@ -188,17 +189,25 @@ def test_involuntary_authorization_decrease(
     threshold_staking.authorizationDecreaseRequested(
         staking_provider, value // 2, 0, sender=creator
```
cygnusv, line 190: Let's try to avoid using `value // 2` in tests, since both the deauthorizing and remaining amount are the same and this can hide test failures.

cygnusv (approved):

cygnusv (approved):

tests/application/test_authorization.py
vzotova: yep, completely agree, all new tests I used something like `value // 3`, I can prepare another PR with updating other test cases

cygnusv (approved):

cygnusv:
It seems I keep forgetting I already approved it 😅

PR #245, opened 2024-02-24
Adds additional role to trigger `withdrawReward`

### LINKED ISSUE #244: Allow MerkleDistribution contract to trigger withdraw TACoApp rewards
Once we introduce rewards in the TACoApp contract, this will produce a fragmentation of the rewards between other apps (tBTC & RB, currently on the MerkleDistribution contract) and TACo. In order to eliminate this friction from a staker perspective, we should allow the MerkleDistribution contract to trigger withdrawing rewards to the beneficiary (currently, only the beneficiary can call this method in TACoApp).

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**Issues fixed/closed:**
Fixes #244

### DIFF
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -46,10 +46,10 @@ contract TACoApplication is
 
     /**
      * @notice Signals that the beneficiary related to the staking provider received reward
+     * @param sender Sender of transaction
      * @param stakingProvider Staking provider address
      * @param beneficiary Beneficiary address
      * @param reward Amount of reward
-     * @param sender Sender of transaction
      */
     event RewardPaid(
         address indexed sender,
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -33,6 +33,12 @@ contract TACoApplication is
     event RewardDistributorSet(address indexed distributor);
 
     /**
+     * @notice Signals that reward contract role was set
+     * @param rewardContract Address of reward contract
+     */
+    event RewardContractSet(address indexed rewardContract);
+
+    /**
      * @notice Signals that reward was added
      * @param reward Amount of reward
      */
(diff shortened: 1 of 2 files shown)


contracts/contracts/TACoApplication.sol
```diff
@@ -43,8 +49,14 @@ contract TACoApplication is
      * @param stakingProvider Staking provider address
      * @param beneficiary Beneficiary address
      * @param reward Amount of reward
+     * @param sender Sender of transaction
```
derekpierre, line 52: Reorder params - `sender` should be first?
```suggestion
     * @param sender Sender of transaction
     * @param stakingProvider Staking provider address
     * @param beneficiary Beneficiary address
     * @param reward Amount of reward
```

contracts/contracts/TACoApplication.sol
```diff
@@ -428,18 +450,21 @@ contract TACoApplication is
...
 
     /**
-     * @notice Withdraw available amount of T reward to beneficiary. Can be called only by beneficiary
+     * @notice Withdraw available amount of T reward to beneficiary. Can be called only by beneficiary or by reward distribution contract
      * @param _stakingProvider Staking provider address
      */
     function withdrawRewards(address _stakingProvider) external updateReward(_stakingProvider) {
         address beneficiary = getBeneficiary(_stakingProvider);
-        require(msg.sender == beneficiary, "Caller must be beneficiary");
+        require(
```
derekpierre, line 458: Out of curiosity - what's the concern if any random entity (not beneficiary or reward contract) calls this function?

vzotova: I can't remember what was exactly underneath, my guess is in case it's tax related operation then someone can troll with it

derekpierre: Ah right - that makes sense 👍 .

derekpierre: I guess to that point - presumably there are restrictions on who can call the corresponding reward contract method (which then calls this method)?

derekpierre (approved):
🎸

manumonti (approved):
LGTM! 👌

theref (approved):

PR #248, opened 2024-03-27
[WIP] Peer Snitching Contract

theref:
**Type of PR:**

- [ ] Feature

**Required reviews:** 
- [X] 3

### DIFF
--- a/contracts/contracts/TACoEvidence.sol
+++ b/contracts/contracts/TACoEvidence.sol
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// Assuming the use of OpenZeppelin contracts for secure cryptographic operations
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "../threshold/ITACoChildApplication.sol";
+
+contract TACoEvidence {
+    using ECDSA for bytes32;
+
+    // Event declarations for logging
+    event CollectionStarted(
+        uint32 indexed id,
+        uint32 indexed startTime,
+        uint32 endTime,
+        uint256 nonce
+    );
+
+    struct Submission {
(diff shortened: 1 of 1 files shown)


contracts/contracts/TACoEvidence.sol
```diff
@@ -0,0 +1,93 @@
...
+        emit CollectionStarted(
+            id,
+            collection.initTimestamp,
+            collection.endTimestamp,
+            collection.nonce
+        );
+    }
+
+    // Function for nodes to submit their online status with evidence
+    function submitEvidence(uint32 id, bytes[] memory signatures, address[] memory peers) public {
```
cygnusv, line 63: I'd remove the `peers` parameter, and simply extract the operator address (as you currently do in L85), and then cross-check with the staking provider address.

contracts/contracts/TACoEvidence.sol
```diff
@@ -0,0 +1,93 @@
...
+    struct Collection {
+        uint32 initTimestamp;
+        uint32 endTimestamp;
+        uint256 nonce;
+        mapping(address => Submission[]) submissions;
+    }
+
+    ITACoChildApplication public immutable application;
+
+    uint256 public submissionWindow = 1 hours;
```
cygnusv, line 32: This can probably be made `immutable`.

contracts/contracts/TACoEvidence.sol
```diff
@@ -0,0 +1,93 @@
...
+    Collection[] public collections;
+
+    constructor() {
+        admin = msg.sender;
```
cygnusv, line 36: This isn't defined, right? Anyway, I'd define as immutable too.

vzotova: I'd change it to Ownable and owner

PR #254, opened 2024-04-12
Penalty for reward as alternative to slashing

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Penalize reward as an option when slashing is too much. Penalty decreases reward by `penaltyDefault` for `penaltyDuration`. Each `penalize` restarts time, penalty itself is the same value until end of duration. Initially this event starts on Polygon in child app, where `penalize` can be called by special role (`adjudicator` for now)

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -173,7 +173,7 @@ contract TACoApplication is
      * @param penaltyPercent Percent of reward that was penalized
      * @param endPenalty End of penalty
      */
-    event Penalized(address indexed stakingProvider, uint256 penaltyPercent, uint256 endPenalty);
+    event Penalized(address indexed stakingProvider, uint256 penaltyPercent, uint64 endPenalty);
 
     /**
      * @notice Signals that reward was reset after penalty
@@ -242,7 +242,7 @@ contract TACoApplication is
      * @param _deauthorizationDuration Duration of decreasing authorization in seconds
      * @param _commitmentDurationOptions Options for commitment duration
      * @param _commitmentDeadline Last date to make a commitment
-     * @param _penaltyDefault Default penalty percentage
+     * @param _penaltyDefault Default penalty percentage (as a value out of 10000)
      * @param _penaltyDuration Duration of penalty
      */
     constructor(
@@ -484,7 +484,9 @@ contract TACoApplication is
         return uint96((_authorized * (PENALTY_BASE - _penaltyPercent)) / PENALTY_BASE);
     }
 
-    /// @dev This view should be called after updateReward modifier
+    /// @dev In case that a penalty period already ended, this view method may produce
+    ///      outdated results if the penalty hasn't been reset, either by calling
+    ///      `resetReward` explicitly or any function with the `updateReward` modifier.
     function effectiveAuthorized(
         uint96 _authorized,
         StakingProviderInfo storage _info
@@ -495,7 +497,9 @@ contract TACoApplication is
         return effectiveAuthorized(_authorized, _info.penaltyPercent);
     }
 
-    /// @dev This view should be called after updateReward modifier
+    /// @dev In case that a penalty period already ended, this view method may produce
+    ///      outdated results if the penalty hasn't been reset, either by calling
+    ///      `resetReward` explicitly or any function with the `updateReward` modifier.
     function effectiveDifference(
         uint96 _from,
         uint96 _to,
@@ -1077,12 +1081,15 @@ contract TACoApplication is
     }
 
     /**
-     * @notice Resets future reward back to 100%
+     * @notice Resets future reward back to 100%.
+     *         Either this method or any method with `updateReward` modifier should be called
+     *         to stop penalties. Otherwise, reward will be still subtracted
+     *         even after the end of penalties.
      * @param _stakingProvider Staking provider address
      */
     function resetReward(address _stakingProvider) external {
         StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
-        require(info.endPenalty != 0, "There are no any penalties");
+        require(info.endPenalty != 0, "There is no penalty");
         require(info.endPenalty <= block.timestamp, "Penalty is still ongoing");
         updateRewardInternal(_stakingProvider);
     }
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -167,25 +167,45 @@ contract TACoApplication is
         address operator
     );
 
+    /**
+     * @notice Signals that the staking provider was penalized
+     * @param stakingProvider Staking provider address
+     * @param penaltyPercent Percent of reward that was penalized
+     * @param endPenalty End of penalty
+     */
+    event Penalized(address indexed stakingProvider, uint256 penaltyPercent, uint256 endPenalty);
+
+    /**
+     * @notice Signals that reward was reset after penalty
+     * @param stakingProvider Staking provider address
+     */
+    event RewardReset(address indexed stakingProvider);
+
     struct StakingProviderInfo {
         address operator;
         bool operatorConfirmed;
         uint64 operatorStartTimestamp;
         uint96 authorized;
-        uint96 deauthorizing; // TODO real usage only in getActiveStakingProviders, maybe remove?
+        uint96 deauthorizing;
         uint64 endDeauthorization;
         uint96 tReward;
         uint160 rewardPerTokenPaid;
         uint64 endCommitment;
+        uint256 stub;
+        uint192 penaltyPercent;
+        uint64 endPenalty;
     }
 
     uint256 public constant REWARD_PER_TOKEN_MULTIPLIER = 10 ** 3;
     uint256 internal constant FLOATING_POINT_DIVISOR = REWARD_PER_TOKEN_MULTIPLIER * 10 ** 18;
+    uint256 public constant PENALTY_BASE = 10000;
 
     uint96 public immutable minimumAuthorization;
     uint256 public immutable minOperatorSeconds;
     uint256 public immutable rewardDuration;
     uint256 public immutable deauthorizationDuration;
+    uint192 public immutable penaltyDefault;
+    uint256 public immutable penaltyDuration;
 
     uint64 public immutable commitmentDurationOption1;
     uint64 public immutable commitmentDurationOption2;
@@ -222,6 +242,8 @@ contract TACoApplication is
      * @param _deauthorizationDuration Duration of decreasing authorization in seconds
      * @param _commitmentDurationOptions Options for commitment duration
      * @param _commitmentDeadline Last date to make a commitment
+     * @param _penaltyDefault Default penalty percentage
+     * @param _penaltyDuration Duration of penalty
      */
     constructor(
         IERC20 _token,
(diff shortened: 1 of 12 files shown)


manumonti (changes requested):

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -54,14 +61,18 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
     /**
      * @notice Initialize function for using with OpenZeppelin proxy
      */
-    function initialize(address _coordinator) external initializer {
-        require(coordinator == address(0), "Coordinator already set");
-        require(_coordinator != address(0), "Coordinator must be specified");
+    function initialize(address _coordinator, address _adjudicator) external initializer {
+        require(coordinator == address(0) || _adjudicator == address(0), "Contracts already set");
```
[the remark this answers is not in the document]
vzotova: of course! thank you for catching it

contracts/contracts/TACoApplication.sol
```diff
@@ -167,25 +167,45 @@ contract TACoApplication is
...
         bool operatorConfirmed;
         uint64 operatorStartTimestamp;
         uint96 authorized;
-        uint96 deauthorizing; // TODO real usage only in getActiveStakingProviders, maybe remove?
+        uint96 deauthorizing;
         uint64 endDeauthorization;
         uint96 tReward;
         uint160 rewardPerTokenPaid;
         uint64 endCommitment;
+        uint256 stub;
```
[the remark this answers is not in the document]
manumonti: now it makes sense

manumonti (approved):
Great work here 🙌

contracts/contracts/TACoApplication.sol
```diff
@@ -222,6 +242,8 @@ contract TACoApplication is
      * @param _deauthorizationDuration Duration of decreasing authorization in seconds
      * @param _commitmentDurationOptions Options for commitment duration
      * @param _commitmentDeadline Last date to make a commitment
+     * @param _penaltyDefault Default penalty percentage
```
derekpierre, line 245: Trying to figure out how to make it known/clear that underlying value should be out of 10000 for calcs.
```suggestion
     * @param _penaltyDefault Default penalty percentage (as a value out of 10000)
```

Unless we want to take a value and adjust it to be out of 10000.

contracts/contracts/TACoApplication.sol
```diff
@@ -957,4 +1051,39 @@ contract TACoApplication is
...
+        emit Penalized(_stakingProvider, info.penaltyPercent, info.endPenalty);
+    }
+
+    /**
+     * @notice Resets future reward back to 100%
+     * @param _stakingProvider Staking provider address
+     */
+    function resetReward(address _stakingProvider) external {
+        StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
+        require(info.endPenalty != 0, "There are no any penalties");
```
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/TACoApplication.sol
```diff
@@ -957,4 +1051,39 @@ contract TACoApplication is
...
+            "Only child application allowed to penalize"
+        );
+
+        if (_stakingProvider == address(0)) {
+            return;
+        }
+
+        StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
+        uint96 before = effectiveAuthorized(info.authorized, info.penaltyPercent);
+        info.endPenalty = uint64(block.timestamp + penaltyDuration);
```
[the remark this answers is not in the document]
vzotova: not really, changed!
✔️

contracts/contracts/TACoApplication.sol
vzotova: ✔️

derekpierre (approved):
🎸 - nice work!

contracts/contracts/TACoApplication.sol
```diff
@@ -957,4 +1051,39 @@ contract TACoApplication is
...
+        info.penaltyPercent = penaltyDefault;
```
arjunhassard, line 1072: Can we add logic to this function such that the `penaltyPercent` is increased by previous violations? To make it as simple as possible:  
- `penaltyDuration` can be a constant (e.g. 1 month for all violations) 
- A violation that occurs during an already penalized reward period simply restarts the 1 month clock from the day that it occurs, but with increased severity. It does not lead to any special outcomes (e.g. summing `penaltyPercent`s) 
- An address's violation count resets after a universal fixed duration (e.g. 24 months), after which point a new violation would result in the least severe `penaltyPercent`. 

Hypothetical parameters: the first incomplete DKG violation leads to a 25% reduction in rewards for a month. The second leads to a 50%, third to a 75%, and fourth (& fifth, sixth...nth) to a 100% reduction. 

Depending on resultant behavior, we can consider increasing `penaltyDuration` for violations beyond the fourth. The rationale for avoiding this in the first version is that we don't want extended periods where nodes are earning nothing, as that may encourage them to exit the network.

arjunhassard (commented):
Nice one! Tried to outline the simplest way for `penaltyPercent` to scale with violation count.

contracts/contracts/TACoApplication.sol
manumonti: This sounds like a good idea. Just one consideration:

- About the penalty duration increasing beyond the fourth: I think we should consider setting its max value to the deauthorization delay (183 days) or less. If this reset duration is much higher than the deauth delay, the [reoffending] operators with 100% of the penalty will be incentivized to just remove the operator from the network and set up a new one with a different staking provider address to avoid the penalty. And this can lead to an unstable network with cohorts with unreachable nodes.

contracts/contracts/TACoApplication.sol
```diff
@@ -167,25 +167,45 @@ contract TACoApplication is
...
         uint64 operatorStartTimestamp;
         uint96 authorized;
-        uint96 deauthorizing; // TODO real usage only in getActiveStakingProviders, maybe remove?
+        uint96 deauthorizing;
         uint64 endDeauthorization;
         uint96 tReward;
         uint160 rewardPerTokenPaid;
         uint64 endCommitment;
+        uint256 stub;
+        uint192 penaltyPercent;
```
cygnusv, line 195: What about using `uint16` for `penaltyPercent`? It leaves some space for later use.

contracts/contracts/TACoApplication.sol
```diff
@@ -957,4 +1051,39 @@ contract TACoApplication is
...
+            authorizedOverall -= before - effectiveAuthorized(info.authorized, info.penaltyPercent);
+        }
+        emit Penalized(_stakingProvider, info.penaltyPercent, info.endPenalty);
+    }
+
+    /**
+     * @notice Resets future reward back to 100%
+     * @param _stakingProvider Staking provider address
+     */
+    function resetReward(address _stakingProvider) external {
```
cygnusv, line 1083: Can you clarify what happens in `resetReward` is not called for a penalized staking provider whose penalty ended?

contracts/contracts/TACoApplication.sol
```diff
@@ -420,12 +470,45 @@ contract TACoApplication is
...
+        uint96 _authorized,
+        StakingProviderInfo storage _info
+    ) internal view returns (uint96) {
+        if (_info.endPenalty == 0) {
+            return _info.authorized;
+        }
+        return effectiveAuthorized(_authorized, _info.penaltyPercent);
+    }
+
+    /// @dev This view should be called after updateReward modifier
```
cygnusv, line 498: What do you mean with these @dev comments?

contracts/contracts/TACoApplication.sol
```diff
@@ -420,12 +470,45 @@ contract TACoApplication is
...
+        uint96 _from,
+        uint96 _to,
+        StakingProviderInfo storage _info
+    ) internal view returns (uint96) {
+        if (_info.endPenalty == 0) {
+            return _from - _to;
+        }
+        uint96 effectiveFrom = effectiveAuthorized(_from, _info.penaltyPercent);
+        uint96 effectiveTo = effectiveAuthorized(_to, _info.penaltyPercent);
+        return effectiveFrom - effectiveTo;
```
cygnusv, line 509: Can you just say `return effectiveAuthorized(_from, _to)`? When `endPenalty == 0`, L492 will make sure the result is `_from - _to`.

contracts/contracts/TACoApplication.sol
cygnusv: @arjunhassard I'm not following completely, can you elaborate a bit further?

contracts/contracts/TACoApplication.sol
vzotova: I'd rather not, penalty percent and duration tight together and updated together so it makes sense to put them in one slot but in general I'd avoid to combine slots just in case, it won't give huge gas savings in advance anyway

contracts/contracts/TACoApplication.sol
vzotova: `resetReward` also included in `updareReward`, so if it none such methods will be called then reward calculation will be looking like staker still has penalty. I know it's not great but with our model - we need active input to update values

contracts/contracts/TACoApplication.sol
vzotova: `updateReward` modifier has `resetReward` inside, so this view will give proper value after that check. Otherwise it will show penalty in cases when they may be reseted

contracts/contracts/TACoApplication.sol
vzotova: not sure that I follow you, second value in `effectiveAuthorized` is penalty percent, can you elaborate? 
in general I made this thing to decrease read operations (still costly)

contracts/contracts/TACoApplication.sol
vzotova: @arjunhassard maybe open this as an issue and I'll address that in separate PR?

cygnusv: https://github.com/nucypher/nucypher-contracts/issues/259

contracts/contracts/TACoApplication.sol
cygnusv: 👍

contracts/contracts/TACoApplication.sol
cygnusv: Right, this is related to my previous comment. What about something like:
```
/// @dev In case that a penalty period already ended, this view method may produce
///      outdated results if the penalty hasn't been reset, either by calling 
///      `resetReward` explicitly or any function with the `updateReward` modifier.
```

contracts/contracts/TACoApplication.sol
cygnusv: Sorry, this is what I meant:
```suggestion
        return effectiveAuthorized(_from - _to);
```

cygnusv (approved):
Added some comments, but in general looks good. Just waiting for the small RFCs to merge, IMO.

contracts/contracts/TACoApplication.sol
vzotova: now I see your question, but I didn't do that on purpose, reason is it could produce different result because of rounding errors, for all other places it's division first and then addition/subtraction.

contracts/contracts/TACoApplication.sol
vzotova: ✔️

contracts/contracts/TACoApplication.sol
vzotova: ✔️

contracts/contracts/TACoApplication.sol
cygnusv: Opened https://github.com/nucypher/nucypher-contracts/issues/260

PR #255, opened 2024-04-15
`ManagedAllowList`

piotr-roslaniec:
**Type of PR:**
- Feature

**Required reviews:** 
- 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**
- Rebased over https://github.com/nucypher/nucypher-contracts/pull/253


contracts/contracts/coordination/SubscriptionManager.sol
```diff
@@ -0,0 +1,118 @@
...
+        require(
+            subscriptions[key].subscriber == address(0),
+            "Subscriber already has an active subscription"
+        );
+
+        Subscription memory newSubscription = Subscription({
+            subscriber: _subscriber,
+            amount: _amount,
+            period: _period
+        });
```
vzotova, line 46: that is a bit ineffective, creating memory variable and then copy it to storage. Would be better just fill up storage variable like in renew method: 
```
Subscription storage newSubscription = subscriptions[key]; 
newSubscription.subscriber = _subscriber;
...
```

piotr-roslaniec:
Hi @cygnusv, @vzotova - I asked for a review of this draft to make sure I'm on the right path with everything. Please let me know what you think.

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -15,63 +19,152 @@ contract GlobalAllowList is IEncryptionAuthorizer {
...
+     * @param ritualId The ID of the ritual
+     * @param addresses The addresses to be authorized or deauthorized
+     * @param value The authorization status
+     */
     function setAuthorizations(uint32 ritualId, address[] calldata addresses, bool value) internal {
-        require(coordinator.isRitualActive(ritualId), "Only active rituals can add authorizations");
+        require(coordinator.isRitualActive(ritualId), "Only active rituals can set authorizations");
+
+        _beforeSetAuthorization(ritualId, addresses, value);
         for (uint256 i = 0; i < addresses.length; i++) {
```
vzotova, line 163: Is it possible to have some cap for number of addresses here? like constant with some not too big value?
any uncapped loop in solidity feels like time bomb

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -0,0 +1,152 @@
...
+     * @param _coordinator The address of the coordinator contract
+     * @param _subscription The address of the subscription contract
+     */
+    constructor(
+        Coordinator _coordinator,
+        UpfrontSubscriptionWithEncryptorsCap _subscription
+    ) GlobalAllowList(_coordinator) {
+        require(address(_coordinator) != address(0), "Coordinator cannot be the zero address");
+        require(address(_subscription) != address(0), "Subscription cannot be the zero address");
+        subscription = _subscription;
```
vzotova, line 44: I do not see any method to change subscription, maybe we can make it `immutable`?

contracts/contracts/coordination/Subscription.sol
```diff
@@ -0,0 +1,261 @@
...
+
+    /**
+     * @notice Returns the key used to lookup authorizations
+     * @param ritualId The ID of the ritual
+     * @param encryptor The address of the encryptor
+     * @return The key used to lookup authorizations
+     */
+    function lookupKey(uint32 ritualId, address encryptor) internal pure returns (bytes32) {
+        return keccak256(abi.encodePacked(ritualId, encryptor));
+    }
```
vzotova, line 116: I suggest to extract this method to library or to just some common contract unless it's too completely different key building here and in allow list

contracts/contracts/coordination/Subscription.sol
```diff
@@ -0,0 +1,261 @@
...
+        uint256 amount = subscriptionFee();
+
+        SubscriptionInfo storage sub = subscriptions[subscriptionId];
+        sub.paidFor += amount;
+        sub.expiration += baseExpiration();
+
+        feeToken.safeTransferFrom(msg.sender, address(this), amount);
+
+        // TODO: We already emit SubscriptionCreated, do we need this?
+        emit SubscriptionPaid(subscriptionId, msg.sender, amount);
```
vzotova, line 148: I'm for keeping this event, because method is public and I suppose it can be called outside without `SubscriptionCreated`

contracts/contracts/coordination/Subscription.sol
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    /**
+     * @notice Withdraws the contract balance to the beneficiary
+     * @param amount The amount to withdraw
+     */
+    function withdrawToBeneficiary(uint256 amount) external {
+        require(msg.sender == beneficiary, "Only the beneficiary can withdraw");
+        uint256 contractBalance = feeToken.balanceOf(address(this));
+        require(contractBalance >= amount, "Insufficient contract balance");
```
vzotova, line 184: I think we can skip this check, because `feeToken` will do that anyway, unless here should be additional rules.
Other question here: if everything withdrawn to beneficiary (btw event would be great) then `cancelSubscription` would not work anymore, is it correct behavior?

contracts/contracts/coordination/Subscription.sol
```diff
@@ -0,0 +1,261 @@
...
+     * @param _feeToken The address of the fee token contract
+     * @param _beneficiary The address of the beneficiary
+     */
+    constructor(Coordinator _coordinator, IERC20 _feeToken, address _beneficiary) {
+        require(address(_coordinator) != address(0), "Coordinator cannot be the zero address");
+        require(address(_feeToken) != address(0), "Fee token cannot be the zero address");
+        require(_beneficiary != address(0), "Beneficiary cannot be the zero address");
+        coordinator = _coordinator;
+        feeToken = _feeToken;
+        beneficiary = _beneficiary;
```
vzotova, line 89: can we make these addresses immutable? at least coordinator and feeToken, unless they can be changed without upgrade in the future

vzotova (commented):
couple high level and style questions:

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -15,63 +19,152 @@ contract GlobalAllowList is IEncryptionAuthorizer {
...
+     * @param ritualId The ID of the ritual
+     * @param encryptor The address of the encryptor
+     * @return The authorization status
+     */
     function isAddressAuthorized(uint32 ritualId, address encryptor) public view returns (bool) {
         return authorizations[lookupKey(ritualId, encryptor)];
     }
 
+    /**
+     * @notice Checks if an address is authorized for a ritual
```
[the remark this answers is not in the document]
piotr-roslaniec: Ah, yes, it's accidental (similar to other comments). I'm not sure what the `@notice` should be for some of these, so I'll keep them blank for the time.

contracts/contracts/coordination/Subscription.sol
piotr-roslaniec: I think in that case we should only withdraw from expired subscriptions, to avoid breaking `cancelSubscription`.

vzotova: that makes sense, so I guess we want some check for that amount?

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -92,9 +92,9 @@ contract ManagedAllowList is GlobalAllowList {
         for (uint256 i = 0; i < addresses.length; i++) {
             // If we want to authorize an address, we need to check if the authorization cap has been exceeded
             // If we want to deauthorize an address, we don't need to check the authorization cap
+            require(!value, "Cannot deauthorize an address");
```
vzotova, line 95: I meant before loop, to not check it each iteration

contracts/contracts/lib/LookupKey.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity ^0.8.0;
+
+library LookupKey {
+    /**
+     * @notice Returns the key used to lookup authorizations
+     * @param ritualId The ID of the ritual
+     * @param encryptor The address of the encryptor
+     * @return The key used to lookup authorizations
+     */
+    function lookupKey(uint32 ritualId, address encryptor) external pure returns (bytes32) {
```
vzotova, line 11: I'd suggest to have it `internal`, some time before there were many troubles with external/public functions in libraries. But if it's working well then no issue

piotr-roslaniec: Setting `internal` here also fixed a library linking issue I had with `ape` :+1:

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -0,0 +1,152 @@
...
+    function _beforeSetAuthorization(
+        uint32 ritualId,
+        address[] calldata addresses,
+        bool value
+    ) internal view override {
+        for (uint256 i = 0; i < addresses.length; i++) {
+            // If we want to authorize an address, we need to check if the authorization cap has been exceeded
+            // If we want to deauthorize an address, we don't need to check the authorization cap
+            require(
+                !value ||
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes. But I'm not sure about the business logic here. Should deauthorizations count towards billable `authActions`? I'm going to assume yes for the time being and remove the `require` on `value`. CC: @cygnusv

contracts/contracts/coordination/Subscription.sol
```diff
@@ -0,0 +1,271 @@
...
+    function withdrawToBeneficiary(uint256 amount) external {
+        require(msg.sender == beneficiary, "Only the beneficiary can withdraw");
+
+        uint256 availableAmount = 0;
+        for (uint32 i = 0; i < numberOfSubscriptions; i++) {
+            SubscriptionInfo storage sub = subscriptions[i];
+            if (block.timestamp >= sub.expiration) {
+                availableAmount += sub.paidFor - sub.spent;
+            }
+        }
```
vzotova, line 190: sorry to nitpick again here, how about to extract this calculation to public view function? it will help to caller to check in advance how much can be withdrawn

piotr-roslaniec:
Requesting review from @arjunhassard and @cygnusv to make sure this matches business logic as outlined in https://github.com/nucypher/nucypher-contracts/issues/257

cygnusv (approved):
Thanks for this @piotr-roslaniec! I think this is a great starting point also for the Subscription stuff laid out in #264

contracts/contracts/coordination/Subscription.sol
```diff
@@ -0,0 +1,279 @@
...
+     * @param subscriptionId The ID of the subscription
+     * @param amount The amount to spend
+     */
+    function spendFromSubscription(uint32 subscriptionId, uint256 amount) external {
+        require(canSpendFromSubscription(subscriptionId, msg.sender), "Unauthorized spender");
+        feeToken.safeTransferFrom(address(this), msg.sender, amount);
+    }
+
+    /**
+     * @notice Calculates the available amount of fees that can be withdrawn be the beneficiary
```
derekpierre, line 178:
```suggestion
     * @notice Calculates the available amount of fees that can be withdrawn by the beneficiary
```

contracts/contracts/lib/LookupKey.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity ^0.8.0;
+
+library LookupKey {
+    /**
+     * @notice Returns the key used to lookup authorizations
+     * @param ritualId The ID of the ritual
+     * @param encryptor The address of the encryptor
+     * @return The key used to lookup authorizations
+     */
+    function lookupKey(uint32 ritualId, address encryptor) internal pure returns (bytes32) {
```
derekpierre, line 11: I don't think `encryptor` applies here anymore since this method is also used for administrator addresses, spender addresses etc. Should we use a more generic name eg. `anAddress` or `associatedAddress`  or ...

scripts/deploy_managed_allow_list.py
```diff
@@ -0,0 +1,26 @@
...
+VERIFY = False
+CONSTRUCTOR_PARAMS_FILEPATH = CONSTRUCTOR_PARAMS_DIR / "managed_allow_list.yml"
+LYNX = ARTIFACTS_DIR / "lynx.json"
+TAPIR = ARTIFACTS_DIR / "tapir.json"
+
+
+def main():
+    deployer = Deployer.from_yaml(filepath=CONSTRUCTOR_PARAMS_FILEPATH, verify=VERIFY)
+    mnaged_allow_list = deployer.deploy(project.ManagedAllowList)
+    deployments = [mnaged_allow_list]
```
derekpierre, line 18:
```suggestion
    managed_allow_list = deployer.deploy(project.ManagedAllowList)
    deployments = [managed_allow_list]
```

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -0,0 +1,151 @@
...
+        address[] calldata addresses,
+        // TODO: Currently unused, remove?
+        // solhint-disable-next-line no-unused-vars
+        bool value
+    ) internal view override {
+        for (uint256 i = 0; i < addresses.length; i++) {
+            require(
+                authActions[ritualId] <
+                    subscription.authorizationActionsCap(ritualId, addresses[i]),
+                "Authorization cap exceeded"
```
derekpierre, line 98: I may be misunderstanding, but this logic is a little unclear to me:
- Should we only be checking caps if `value` is `True` i.e. we are authorizing? If deauthorizing (i.e. value is False), there are no caps.
- I'm unclear what we are checking in the `require` statement. `canSetAuthorizations` already checks whether the sender is a valid admin for the ritual - great, but I assumed here you need to check 2 things:
   1. the remaining allowance for the admin making the authorization request (i.e. sender) allows for the set of encryptor addresses to be authorized, so:
   ```solidity
   require("getAllowance(ritualId, msg.sender) >= addresses.length", "Administrator cap exceeded")
   ```
   2. That the total cap across all administrators for a specific ritual is not being exceeded, so:
   ```solidity
   require("authActions[ritualId] >= addresses.length", "Ritual-wide cap exceeded")
   ```

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -15,63 +20,143 @@ contract GlobalAllowList is IEncryptionAuthorizer {
...
+        require(addresses.length <= MAX_AUTH_ACTIONS, "Too many addresses");
+
+        _beforeSetAuthorization(ritualId, addresses, value);
         for (uint256 i = 0; i < addresses.length; i++) {
-            authorizations[lookupKey(ritualId, addresses[i])] = value;
+            authorizations[LookupKey.lookupKey(ritualId, addresses[i])] = value;
             emit AddressAuthorizationSet(ritualId, addresses[i], value);
         }
+
+        authActions[ritualId] += addresses.length;
```
derekpierre, line 160: Is incrementing only applicable if `value` is `True`? If `value` is `False` should this be decrementing instead?

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -0,0 +1,151 @@
...
+    ) internal {
+        require(
+            coordinator.isRitualActive(ritualId),
+            "Only active rituals can set administrator caps"
+        );
+        for (uint256 i = 0; i < addresses.length; i++) {
+            allowance[LookupKey.lookupKey(ritualId, addresses[i])] = value;
+            emit AdministratorCapSet(ritualId, addresses[i], value);
+        }
+        authActions[ritualId] += addresses.length;
```
derekpierre, line 123: Incrementing if `value` is `True`. Should we be decrementing if `value` is `False`?

contracts/contracts/coordination/GlobalAllowList.sol
cygnusv: For this version, the goal was to charge per action, regardless if it's authorizing or deauthorizing. But more recent variations proposed to some adopters will use something different (e.g. charging for authorized slots). I think we can leave it as is because it's only a starting point anyway, and the more recent design of subscription contracts will refactor everything

contracts/contracts/coordination/ManagedAllowList.sol
derekpierre: My calls may be wrong above ^, but the pseudocode I'm looking for is:
i. remaining allowance for administrator to be >= the number of encryptor addresses to authorize
ii. remaining actions for ritual >= the number of encryptor addresses to authorize

contracts/contracts/coordination/GlobalAllowList.sol
derekpierre: 👍 I see, so every action (authorize or deauthorize) is a charge.

contracts/contracts/lib/LookupKey.sol
arjunhassard: It's true it's not strictly accurate here, but the hierarchy of roles/names is getting quite unwieldy. Maybe we can rename `encryptor` --> `encryptAdmin` and then rename this `admin` or `anAdmin`.

arjunhassard (commented):
Excellent work, think you can claim to be at least somewhat-versed in Solidity now. 

Still quite a few things to clarify and decide, particularly with respect to #253-driven refactoring. I suggest we do a PR walkthrough, maybe on Friday?

contracts/contracts/lib/LookupKey.sol
piotr-roslaniec: Since this is a library code and is re-used in a couple of places, perhaps we should rename it to just `_address`. `encryptor` and `admin` are terms from specific contexts.

contracts/contracts/coordination/ManagedAllowList.sol
piotr-roslaniec: Just to give more context, I wasn't sure how to use `value` here: Should setting authorization to `false` (so, removing authorization) count towards billable actions? We may treat adding and removing authorization differently, but I think it's application-specific (the frequency of actions, but also the application-specific meaning of authorizing/deauthorizing)

contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -15,63 +20,143 @@ contract GlobalAllowList is IEncryptionAuthorizer {
...
+     * @dev Only active rituals can set authorizations
+     * @param ritualId The ID of the ritual
+     * @param addresses The addresses to be authorized or deauthorized
+     * @param value The authorization status
+     */
     function setAuthorizations(uint32 ritualId, address[] calldata addresses, bool value) internal {
-        require(coordinator.isRitualActive(ritualId), "Only active rituals can add authorizations");
+        require(coordinator.isRitualActive(ritualId), "Only active rituals can set authorizations");
+
+        require(addresses.length <= MAX_AUTH_ACTIONS, "Too many addresses");
```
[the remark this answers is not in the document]
piotr-roslaniec: Yes, per one `setAuthorizations` call

cygnusv:
@arjunhassard @derekpierre The work @vzotova is doing on #264 will be based directly on this (as you can see in https://github.com/nucypher/nucypher-contracts/pull/265, Vicky is already working on top of this PR). The plan we laid out is that this is good foundational work, but that will require a lot of changes in light of the recent design for Subscriptions and the newer requirements for the AllowList contract. I propose we merge this now, either as-is or with minor changes (like renames).

PR #261, opened 2024-05-03
Increment penalty in case of repeated violation during previous duration

### LINKED ISSUE #259: penaltyPercent to scale with each violation
Add logic to TACoApplication.sol such that the `penaltyPercent` parameter increases with every violation committed by a given staker. 

To make it as simple as possible in the first version:  
- All incomplete DKG rituals are treated as equal violations regardless of the reason or context. In future versions there could be varying punishment/severity for: 
   - Incomplete DKGs due to being offline vs. not running an up-to-date version vs. other known failures 
   - Incomplete DKGs where more than one node is affected and/or there is correlation (e.g. cloud provider goes down). Note this could lead to more severe punishments to encourage diverse infrastructure. 
   - Incomplete DKGs that indicate malicious behavior 
   - Incomplete handover DKGs during deauth periods. 
- `penaltyDuration` should be a constant 
   - e.g. 1 month of reduced rewards for all violations
- A violation that occurs during an already penalized reward period simply restarts the 1 month clock from the day the violation is logged (but with increased severity). In other words, committing a violation before a previous punishment has ended does not lead to any special outcomes, like summing `penaltyPercent`s. 
- A staker's violation count resets after a universal fixed duration (e.g. 24 months) from their first violation, after which point a fresh violation would result in the least severe `penaltyPercent`. 

Hypothetical initial parameters: 
- First incomplete DKG violation leads to a 25% reduction in rewards for a month. 
- The second leads to a 50%
- Third to a 75%, 
- Fourth (& fifth, sixth...nth) to a 100% reduction. 

Depending on resultant behavior, we can consider increasing `penaltyDuration` for violations beyond the fourth. The rationale for avoiding this in the first version is that we don't want extended periods where nodes are earning nothing, as that may encourage them to exit the network. 

Comment from @manumonti: 
About the penalty duration increasing beyond the fourth: I think we should consider setting its max value to the deauthorization delay (183 days) or less. If this reset duration is much higher than the deauth delay, the [reoffending] operators with 100% of the penalty will be incentivized to just remove the operator from the network and set up a new one with a different staking provider address to avoid the penalty. And this can lead to an unstable network with cohorts with unreachable nodes.

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
See #259

**Issues fixed/closed:**
Fixes #259 

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -245,7 +245,7 @@ contract TACoApplication is
      * @param _commitmentDeadline Last date to make a commitment
      * @param _penaltyDefault Default penalty percentage (as a value out of 10000)
      * @param _penaltyDuration Duration of penalty
-     * @param _penaltyIncrement Increment of penalty if it happens during previous penalty duration
+     * @param _penaltyIncrement Increment of penalty if violation occurs during an existing penalty period
      */
     constructor(
         IERC20 _token,
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -198,7 +198,7 @@ contract TACoApplication is
 
     uint256 public constant REWARD_PER_TOKEN_MULTIPLIER = 10 ** 3;
     uint256 internal constant FLOATING_POINT_DIVISOR = REWARD_PER_TOKEN_MULTIPLIER * 10 ** 18;
-    uint256 public constant PENALTY_BASE = 10000;
+    uint192 public constant PENALTY_BASE = 10000;
 
     uint96 public immutable minimumAuthorization;
     uint256 public immutable minOperatorSeconds;
@@ -206,6 +206,7 @@ contract TACoApplication is
     uint256 public immutable deauthorizationDuration;
     uint192 public immutable penaltyDefault;
     uint256 public immutable penaltyDuration;
+    uint192 public immutable penaltyIncrement;
 
     uint64 public immutable commitmentDurationOption1;
     uint64 public immutable commitmentDurationOption2;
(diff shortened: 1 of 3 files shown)


manumonti (approved):
LGTM! 👌

derekpierre (approved):
🎸

contracts/contracts/TACoApplication.sol
```diff
@@ -244,6 +245,7 @@ contract TACoApplication is
      * @param _commitmentDeadline Last date to make a commitment
      * @param _penaltyDefault Default penalty percentage (as a value out of 10000)
      * @param _penaltyDuration Duration of penalty
+     * @param _penaltyIncrement Increment of penalty if it happens during previous penalty duration
```
arjunhassard, line 248:
```suggestion
     * @param _penaltyIncrement Increment of penalty if violation occurs during an existing penalty period
```
Just to clarify that 'previous' penalties have no bearing on this variable

arjunhassard (approved):
Nice work! One future-proofing comment; at some point it would be useful to add a variable – `violationTimeframe` or something, which can be tuned so that if there are lots of DKG incompletions within a short timeframe (e.g. 3 hours), the guilty staker is given a reprieve from penalty increments, and only sees their rewards reduced by the default (unless they were already within a penalty period, of course!).

cygnusv (approved):

cygnusv (approved):

Issue #266, opened 2024-05-15
InfractionCollector contract 


cygnusv:
* New contract called `InfractionCollector` (Feel free to rename it, btw!)
* For the moment, ownable and upgradeable.
* `InfractionCollector` lives in Polygon, alongside `Coordinator` and `TACoChildApplication`
* Replace reference to `Adjudicator` in `TACoChildApplication` with `InfractionCollector`
* `InfractionCollector` will have a reference to `Coordinator` and will have a permissionless method `reportDKGInfraction`, with parameters a ritual ID and an array of staking provider addresses:
   * The method will check for unambiguous faults in the given DKG ritual that are attributable to the specific providers.
   * For example, DKG that timed out due to failure to post transactions during the transcript phase.
   * Other cases like aggregation mismatches or failures to post aggregations will require more complex verification, so let's start with this, but let's continue to think other faults that can be unambiguously attributed to specific providers.
   * For all notified infractions, the method will call `TACoChildApplication.penalize(stakingProviderAddress)`. See:
https://github.com/nucypher/nucypher-contracts/blob/449ec475c433043d53bff2b55eaab91d9d71c60e/contracts/contracts/coordination/TACoChildApplication.sol#L202
   * Reported infractions will be marked down in `InfractionCollector` so they can't be reported twice.

PR #267, opened 2024-05-15
InfractionCollector Contract

theref:
Scope is outlined in #266 
- [x] First Draft of Contract
- [x] Updates to existing contracts
- [x] Deployment scripts
- [x] Tests

Related to https://github.com/nucypher/sprints/issues/35

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,12 @@ def main():
     deployer.finalize(deployments=deployments)
 ```
 
+To run a deployment locally:
+
+```bash
+ape run lynx deploy_infraction --network ethereum:local:test
+```
+
 ##### 3. Setup Deployment Account (production only)
 
 In order to deploy to **production** you will need to import an account into ape:
--- a/README.md
+++ b/README.md
@@ -117,12 +117,6 @@ def main():
     deployer.finalize(deployments=deployments)
 ```
 
-To run a deployment locally:
-
-```bash
-ape run lynx deploy_infraction --network ethereum:local:test
-```
-
 ##### 3. Setup Deployment Account (production only)
 
 In order to deploy to **production** you will need to import an account into ape:
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -20,10 +20,7 @@ contract InfractionCollector is OwnableUpgradeable {
     // Mapping to keep track of reported infractions
     mapping(uint32 => mapping(address => mapping(InfractionType => bool))) public infractions;
 
-    constructor(
-        Coordinator _coordinator,
-        ITACoChildApplication _tacoChildApplication
-    ) {
+    constructor(Coordinator _coordinator, ITACoChildApplication _tacoChildApplication) {
         coordinator = _coordinator;
         tacoChildApplication = _tacoChildApplication;
     }
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -7,6 +7,7 @@ import "./Coordinator.sol";
 import "../../threshold/ITACoChildApplication.sol";
 
 contract InfractionCollector is OwnableUpgradeable {
+    event InfractionReported(uint32 indexed ritualId, address indexed stakingProvider, InfractionType infractionType);
     Coordinator public immutable coordinator;
 
     // Reference to the TACoChildApplication contract
@@ -57,6 +58,7 @@ contract InfractionCollector is OwnableUpgradeable {
                 infractions[ritualId][stakingProviders[i]][
                     InfractionType.MISSING_TRANSCRIPT
                 ] = true;
+                emit InfractionReported(ritualId, stakingProviders[i], InfractionType.MISSING_TRANSCRIPT);
             }
         }
     }
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -7,7 +7,11 @@ import "./Coordinator.sol";
 import "../../threshold/ITACoChildApplication.sol";
 
 contract InfractionCollector is OwnableUpgradeable {
-    event InfractionReported(uint32 indexed ritualId, address indexed stakingProvider, InfractionType infractionType);
+    event InfractionReported(
+        uint32 indexed ritualId,
+        address indexed stakingProvider,
+        InfractionType infractionType
+    );
     Coordinator public immutable coordinator;
 
     // Reference to the TACoChildApplication contract
@@ -58,7 +62,11 @@ contract InfractionCollector is OwnableUpgradeable {
                 infractions[ritualId][stakingProviders[i]][
                     InfractionType.MISSING_TRANSCRIPT
                 ] = true;
-                emit InfractionReported(ritualId, stakingProviders[i], InfractionType.MISSING_TRANSCRIPT);
+                emit InfractionReported(
+                    ritualId,
+                    stakingProviders[i],
+                    InfractionType.MISSING_TRANSCRIPT
+                );
             }
         }
     }
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -41,7 +41,7 @@ contract InfractionCollector is OwnableUpgradeable {
                 ritualId,
                 stakingProviders[i]
             );
-            if (participant.transcript.length == 0) {
+            if (participant.transcript.length == 0) {  // Transcript TX wasn't posted
                 // Penalize the staking provider
                 tacoChildApplication.penalize(stakingProviders[i]);
                 infractions[ritualId][stakingProviders[i]] = true;
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -23,7 +23,7 @@ contract InfractionCollector is OwnableUpgradeable {
 
     constructor(Coordinator _coordinator, ITACoChildApplication _tacoChildApplication) {
         require(
-            _coordinator != address(0) && _tacoChildApplication != address(0),
+            address(_coordinator) != address(0) && address(_tacoChildApplication) != address(0),
             "Contracts must be specified"
         );
         coordinator = _coordinator;
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -21,8 +21,13 @@ contract InfractionCollector is OwnableUpgradeable {
     mapping(uint32 => mapping(address => mapping(InfractionType => bool))) public infractions;
 
     constructor(Coordinator _coordinator, ITACoChildApplication _tacoChildApplication) {
+        require(
+            _coordinator != address(0) && _tacoChildApplication != address(0),
+            "Contracts must be specified"
+        );
         coordinator = _coordinator;
         tacoChildApplication = _tacoChildApplication;
+        _disableInitializers();
     }
 
     function reportMissingTranscript(
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -1,4 +1,5 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
 pragma solidity ^0.8.0;
 
 import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -24,10 +24,7 @@ contract InfractionCollector is OwnableUpgradeable {
         public infractionsForRitual;
 
     constructor(Coordinator _coordinator) {
-        require(
-            address(_coordinator) != address(0),
-            "Contracts must be specified"
-        );
+        require(address(_coordinator) != address(0), "Contracts must be specified");
         coordinator = _coordinator;
         tacoChildApplication = coordinator.application();
         _disableInitializers();
@@ -50,7 +47,9 @@ contract InfractionCollector is OwnableUpgradeable {
         for (uint256 i = 0; i < stakingProviders.length; i++) {
             // Check if the infraction has already been reported
             require(
-                infractionsForRitual[ritualId][stakingProviders[i]][InfractionType.MISSING_TRANSCRIPT] == 0,
+                infractionsForRitual[ritualId][stakingProviders[i]][
+                    InfractionType.MISSING_TRANSCRIPT
+                ] == 0,
                 "Infraction already reported"
             );
             Coordinator.Participant memory participant = coordinator.getParticipantFromProvider(
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -13,7 +13,7 @@ contract InfractionCollector is OwnableUpgradeable {
     TACoChildApplication public tacoChildApplication;
 
     // Mapping to keep track of reported infractions
-    mapping(uint32 => mapping(address => bool)) private infractions;
+    mapping(uint32 => mapping(address => bool)) public infractions;
 
     function initialize(
         Coordinator _coordinator,
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -7,10 +7,10 @@ import "./Coordinator.sol";
 import "../../threshold/ITACoChildApplication.sol";
 
 contract InfractionCollector is OwnableUpgradeable {
-    Coordinator public coordinator;
+    Coordinator public immutable coordinator;
 
     // Reference to the TACoChildApplication contract
-    ITACoChildApplication public tacoChildApplication;
+    ITACoChildApplication public immutable tacoChildApplication;
 
     // infraction types
     enum InfractionType {
@@ -20,11 +20,10 @@ contract InfractionCollector is OwnableUpgradeable {
     // Mapping to keep track of reported infractions
     mapping(uint32 => mapping(address => mapping(InfractionType => bool))) public infractions;
 
-    function initialize(
+    constructor(
         Coordinator _coordinator,
         ITACoChildApplication _tacoChildApplication
-    ) public initializer {
-        __Ownable_init(msg.sender);
+    ) {
         coordinator = _coordinator;
         tacoChildApplication = _tacoChildApplication;
     }
--- a/contracts/contracts/coordination/InfractionCollector.sol
+++ b/contracts/contracts/coordination/InfractionCollector.sol
@@ -4,20 +4,20 @@ pragma solidity ^0.8.0;
 
 import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
 import "./Coordinator.sol";
-import "./TACoChildApplication.sol";
+import "../../threshold/ITACoChildApplication.sol";
 
 contract InfractionCollector is OwnableUpgradeable {
     Coordinator public coordinator;
 
     // Reference to the TACoChildApplication contract
-    TACoChildApplication public tacoChildApplication;
+    ITACoChildApplication public tacoChildApplication;
 
     // Mapping to keep track of reported infractions
     mapping(uint32 => mapping(address => bool)) public infractions;
 
     function initialize(
         Coordinator _coordinator,
-        TACoChildApplication _tacoChildApplication
+        ITACoChildApplication _tacoChildApplication
     ) public initializer {
         __Ownable_init(msg.sender);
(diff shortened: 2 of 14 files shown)


contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,50 @@
...
+
+import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import "./Coordinator.sol";
+import "./TACoChildApplication.sol";
+
+contract InfractionCollector is OwnableUpgradeable {
+    Coordinator private coordinator;
+
+    // Reference to the TACoChildApplication contract
+    TACoChildApplication private tacoChildApplication;
```
cygnusv, line 12:
```suggestion
    Coordinator public coordinator;

    // Reference to the TACoChildApplication contract
    TACoChildApplication public tacoChildApplication;
```

cygnusv (commented):
Looking good!

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,51 @@
...
+
+import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
+import "./Coordinator.sol";
+import "./TACoChildApplication.sol";
+
+contract InfractionCollector is OwnableUpgradeable {
+    Coordinator public coordinator;
+
+    // Reference to the TACoChildApplication contract
+    TACoChildApplication public tacoChildApplication;
```
vzotova, line 13: Minor thing, but would be a bit better to use `ITACoChildApplication` interface, it has `penalize` function

theref: Fixed in [Use interface for TACo Child Application in Infraction Collector](https://github.com/nucypher/nucypher-contracts/pull/267/commits/eea7b4a80ec964ee1f6a2d4d836fe65a3d24ec9e)

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    // Mapping to keep track of reported infractions
+    mapping(uint32 => mapping(address => bool)) public infractions;
+
+    function initialize(
+        Coordinator _coordinator,
+        ITACoChildApplication _tacoChildApplication
+    ) public initializer {
+        __Ownable_init(msg.sender);
+        coordinator = _coordinator;
```
derekpierre, line 23: Not sure of best practice in solidity, but should we add some `require` statements before setting  anything? (cc @vzotova )
1. Non-zero addresses
```solidity
require(_coordinator != address(0) && _tacoChildApplication != address(0), "Contracts must be specified");
```
2. Check that the values are not already set
```solidity
require(coordinator == address(0) || tacoChildApplication == address(0), "Contracts already set");
```

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,51 @@
...
+    ) external {
+        // Ritual must have failed
+        require(
+            coordinator.getRitualState(ritualId) == Coordinator.RitualState.DKG_TIMEOUT,
+            "Ritual must have failed"
+        );
+
+        for (uint256 i = 0; i < stakingProviders.length; i++) {
+            // Check if the infraction has already been reported
+            require(!infractions[ritualId][stakingProviders[i]], "Infraction already reported");
```
derekpierre, line 39: Given that we are going through a list of addresses, should we use an `if` instead of a `require`?

deployment/constructor_params/lynx/infraction.yml
```diff
@@ -0,0 +1,18 @@
...
+constants:
+  # See deployment/artifacts/lynx.json
+  COORDINATOR_PROXY: "0xE9e94499bB0f67b9DBD75506ec1735486DE57770"
+  TACO_CHILD_PROXY: "0x42F30AEc1A36995eEFaf9536Eb62BD751F982D32"
+
+contracts:
+  - InfractionCollector:
+      proxy:
+        constructor:
+          _data: $encode:initialize,$COORDINATOR_PROXY,$TACO_CHILD_PROX
```
derekpierre, line 18:
```suggestion
          _data: $encode:initialize,$COORDINATOR_PROXY,$TACO_CHILD_PROXY
```

derekpierre: ^ In case testing was an issue you can test this using a local deployment (this should have failed locally). Happy to do a joint session with you about testing it.

tests/test_infraction.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@pytest.fixture()
+def erc20(project, initiator):
+    token = project.TestToken.deploy(ERC20_SUPPLY, sender=initiator)
+    return token
+
+
+@pytest.fixture()
+def coordinator(project, deployer, application, erc20, initiator, oz_dependency):
```
derekpierre, line 90: I believe this fixture is repeated in `test_coordinator`. More generally, other fixtures are being duplicated as well, e.g., `erc20(...)`, etc. You could consider putting common fixtures into the conftest.py to prevent duplicating code.

deployment/constructor_params/lynx/infraction.yml
derekpierre: Yep. Just ping me.

tests/test_infraction.py
theref: they're all copied over from `test_coordinator`. I would expect all fixtures to be in `conftest` so I assumed there was a reason that were being kept in the same file as the tests that used them. If that's not the case, then I'm very much on board with moving everything to `conftest` and getting rid of all the duplicates

derekpierre: In some cases, a mock contract is used instead of the actual contract because of specific testing needs. For example, see https://github.com/nucypher/nucypher-contracts/blob/main/tests/test_child_application.py#L56 for the `Coordinator` contract.

But other than those nuanced cases, moving common/generic fixtures to `conftest.py` should be fine and overridden as necessary in the individual test modules themselves, like the above.

To keep things simple, this modification does not need to occur in this PR; it could be done in a follow-up PR.

README.md
```diff
@@ -117,6 +117,12 @@ def main():
     deployer.finalize(deployments=deployments)
 ```
 
+To run a deployment locally:
+
+```bash
+ape run lynx deploy_infraction --network ethereum:local:test
```
cygnusv, line 123: I guess this shouldn't be part of the readme, right?

contracts/contracts/coordination/InfractionCollector.sol
cygnusv: Bump on the RFCs from @derekpierre and @vzotova

contracts/contracts/coordination/InfractionCollector.sol
cygnusv: Unsure about this, I see benefits on both sides, but I'd lean towards reverting the TX. 
BTW, I think there should be an `InfractionCollection` specific event, something like `ReportedMissingTranscript` or similar.

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,51 @@
...
+            Coordinator.Participant memory participant = coordinator.getParticipantFromProvider(
+                ritualId,
+                stakingProviders[i]
+            );
+            if (participant.transcript.length == 0) {
```
cygnusv, line 44:
```suggestion
            if (participant.transcript.length == 0) {  // Transcript TX wasn't posted
```

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -61,18 +61,21 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
     /**
      * @notice Initialize function for using with OpenZeppelin proxy
      */
-    function initialize(address _coordinator, address _adjudicator) external initializer {
-        require(coordinator == address(0) || adjudicator == address(0), "Contracts already set");
```
cygnusv, line 65: @vzotova why the OR here? Shouldn't this have an AND?

README.md
theref: how come? I wasn't sure how to do it and @derekpierre explained, so I thought I'd add it in for next time :D

contracts/contracts/coordination/InfractionCollector.sol
theref: @vzotova why move them to constructor? i'm never sure of the best patterns here - especially because tacoChild will need to reference InfractionCollector as well

README.md
cygnusv: The README is currently explaining how deployment scripts work generically, so an instruction that's specific to `lynx deploy_infraction` doesn't quite fit.

derekpierre: There is an instruction in the `Deploy` section - https://github.com/nucypher/nucypher-contracts/blob/fdcb313c81102ab5d52e673dd3e9be80caf9f85b/README.md#4-deploy - that shows the command for using the local network and is more generic, and perhaps already achieves what you wanted to convey. Of course, you can reword/move things around for better clarity. 

Keeping it general is a good thing, and then optionally including specific examples if more clarity is needed.

contracts/contracts/coordination/InfractionCollector.sol
vzotova: when infraction collector will be deployed - we will have taco child and coordinator already working, that's why it's not in taco child constructor, but would be good idea to have it in constructor for collector (it will work as constant which will save gas)

contracts/contracts/coordination/TACoChildApplication.sol
vzotova: because in mainnet coordinator already initialized, so allows to use this function if previous version was deployed before

tests/test_infraction.py
theref: opened an issue to track this https://github.com/nucypher/nucypher-contracts/issues/284

contracts/contracts/coordination/InfractionCollector.sol
theref: fixed in https://github.com/nucypher/nucypher-contracts/pull/267/commits/128ec8e0478197c6acdafb2abdf6c6672427e618

contracts/contracts/coordination/InfractionCollector.sol
theref: fixed in https://github.com/nucypher/nucypher-contracts/pull/267/commits/3bdcc8dcf613489042ff14aa9e3eb460e18a4270

tests/test_child_application.py
```diff
@@ -331,7 +331,7 @@ def test_penalize(accounts, root_application, child_application, coordinator):
     ) = accounts[0:]
 
     # Penalize can be done only from adjudicator address
```
derekpierre, line 333:
```suggestion
    # Penalize can be done only from infraction collector address
```

tests/test_infraction.py
```diff
@@ -0,0 +1,180 @@
...
+    return proxy_contract
+
+
+def test_no_infractions(
+    erc20, nodes, initiator, global_allow_list, infraction_collector, coordinator
+):
+    for node in nodes:
+        public_key = gen_public_key()
+        coordinator.setProviderPublicKey(public_key, sender=node)
+        cost = coordinator.getRitualInitiationCost(nodes, DURATION)
```
derekpierre, line 135: Should this line for `cost` be outside of the `for` loop? It isn't node-specific.

Same comment for other `cost = ...` lines in subsequent tests.

tests/test_infraction.py
```diff
@@ -0,0 +1,180 @@
...
+    erc20.approve(coordinator.address, cost, sender=initiator)
+    coordinator.initiateRitual(
+        nodes, initiator, DURATION, global_allow_list.address, sender=initiator
+    )
+    chain.pending_timestamp += TIMEOUT * 2
+    infraction_collector.reportMissingTranscript(0, nodes, sender=initiator)
+    for node in nodes:
+        assert infraction_collector.infractions(0, node) is True
+
```
derekpierre, line 163: We should also test partial submissions i.e. some nodes submitted transcripts and some did not.

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,62 @@
...
+    // Reference to the TACoChildApplication contract
+    ITACoChildApplication public tacoChildApplication;
+
+    // infraction types
+    enum InfractionType {
+        MISSING_TRANSCRIPT
+    }
+
+    // Mapping to keep track of reported infractions
+    mapping(uint32 => mapping(address => mapping(InfractionType => bool))) public infractions;
```
KPrasch, line 21: What is the reason why the top-level index here is the ritual ID?  I'm not suggesting any changes at this moment but how this data structure is formed has implications on how cost-effective and code-efficient looking up infractions is later.

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,62 @@
...
+                ritualId,
+                stakingProviders[i]
+            );
+            if (participant.transcript.length == 0) {
+                // Transcript TX wasn't posted
+                // Penalize the staking provider
+                tacoChildApplication.penalize(stakingProviders[i]);
+                infractions[ritualId][stakingProviders[i]][
+                    InfractionType.MISSING_TRANSCRIPT
+                ] = true;
```
KPrasch, line 58: Related to my above comment - it's worth considering an alternate structure here so that it's easier to lookup all infractions pertaining to a single node.  Without any real world usage data, it's hard to know exactly what types of on-chain queries we will need to handle penalties later.  Based on prior discussions, we may need to identify multiple subsequent infractions in order to fairly enforce a downtime penalty.

contracts/contracts/coordination/InfractionCollector.sol
vzotova: small question, I do not see commit 128ec8e0478197c6acdafb2abdf6c6672427e618 here. Am I missing it or it was lost?

contracts/contracts/coordination/InfractionCollector.sol
vzotova: taco app is checking subsequent penalties, so we can put it there, otherwise structure will change to mapping and arrays and will consume more gas (not a end of the world though)

deployment/constructor_params/lynx/infraction.yml
```diff
@@ -0,0 +1,19 @@
...
+  # See deployment/artifacts/lynx.json
+  COORDINATOR_PROXY: "0xE9e94499bB0f67b9DBD75506ec1735486DE57770"
+  TACO_CHILD_PROXY: "0x42F30AEc1A36995eEFaf9536Eb62BD751F982D32"
+
+contracts:
+  - InfractionCollector:
+      proxy:
+      constructor:
+          _coordinator: $COORDINATOR_PROXY
+          _tacoChildApplication: $TACO_CHILD_PROXY
```
derekpierre, line 19: 2-space tabbing for yml file. May have originally been my fault with my original code suggestion 😅 

```suggestion
      constructor:
        _coordinator: $COORDINATOR_PROXY
        _tacoChildApplication: $TACO_CHILD_PROXY
```

scripts/lynx/deploy_infraction.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/python3
```
derekpierre, line 1: Did you want to create one of these for `mainnet` and `tapir`? Or, were you thinking of doing that at the same time as actual `mainnet` deployment (which is fine)?

tests/test_infraction.py
```diff
@@ -0,0 +1,170 @@
...
+        "NON_INITIATED",
+        "DKG_AWAITING_TRANSCRIPTS",
+        "DKG_AWAITING_AGGREGATIONS",
+        "DKG_TIMEOUT",
+        "DKG_INVALID",
+        "ACTIVE",
+        "EXPIRED",
+    ],
+    start=0,
+)
```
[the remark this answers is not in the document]
derekpierre: ^ Any thoughts on this @theref ? It would be nice to have it as a common constant.

tests/test_infraction.py
derekpierre: ^ One outstanding comment here. Something like half of the nodes submitted transcripts, but the other half did not.

theref:
@derekpierre not sure why but it won't let me respond directly to your comments.

>  Any thoughts on this @theref ? It would be nice to have it as a common constant.

You and I discussed the fact that A LOT of the testing code is being duplicated. This issue https://github.com/nucypher/nucypher-contracts/issues/284 was opened and we should add it to a sprint in the future, it would be a great piece of cleanup

> One outstanding comment here. Something like half of the nodes submitted transcripts, but the other half did not.

I actually have no idea how to test this :D

derekpierre:
> You and I discussed the fact that A LOT of the testing code is being duplicated. This issue #284 was opened and we should add it to a sprint in the future, it would be a great piece of cleanup

Ah ok. It'll be addressed as part of #284 - 👍

contracts/contracts/coordination/InfractionCollector.sol
derekpierre: > how this data structure is formed has implications on how cost-effective and code-efficient looking up infractions is later.

This is a really interesting consideration.

I don't have clear thoughts on this at the moment, but is it more likely that infractions would be looked up/checked via staking provider address instead of by ritual id? This would be from a dashboard or ...(?). 

For example, would we want to know all infractions for a specific staker address as opposed to all infractions from a failed ritual id? Even in the latter case (all infractions from failed ritual id), it is easy to get the list of all staking provider addresses for a ritual id from the coordinator (which this contract also has a reference to), and then check infractions for each staking provider address.

theref:
@cygnusv @vzotova 
after deploying to lynx @KPrasch and I noticed that the owner of this contract is `0x0...00` address. I can see in other contracts that we do:
```
    function initialize() external initializer {
        __Ownable_init(msg.sender);
    }
```

but that contract also has a `_disableInitializers()` call at the end of it's constructor. I'm a bit confused :) what should we be doing here?

cygnusv:
> @cygnusv @vzotova after deploying to lynx @KPrasch and I noticed that the owner of this contract is `0x0...00` address. I can see in other contracts that we do:
> 
> ```
>     function initialize() external initializer {
>         __Ownable_init(msg.sender);
>     }
> ```
> 
> but that contract also has a `_disableInitializers()` call at the end of it's constructor. I'm a bit confused :) what should we be doing here?

Good question! We need to add the initialize function you suggest, and also keep the disableInitializers() call. The former has effect on the proxy state (which is what proxies are for) while the latter disables initializers on logic contracts.

KPrasch (commented):
Great work breaking the ice on this and getting it deployed onto testnets.  Let's spend some time understanding how to implement the penalty design, collect more data on testnet, and use that information to continue development on this contract.  

That being said, I think it's wise to open a new epic branch on the upstream remote so that we can properly accept these changes with a merge.

KPrasch (approved):
I changed the base to the `infractions` branch -- Approved on that basis 👍🏻

tests/test_infraction.py
```diff
@@ -0,0 +1,183 @@
...
+    # post transcript for half of nodes
+    for node in nodes[:len(nodes) // 2]:
+        coordinator.postTranscript(0, transcript, sender=node)
+    chain.pending_timestamp += TIMEOUT * 2
+    infraction_collector.reportMissingTranscript(0, nodes, sender=initiator)
+    # first half of nodes should be fine, second half should be infracted
+    for node in nodes[:len(nodes) // 2]:
+        assert infraction_collector.infractions(0, node, 0) == False
+    for node in nodes[len(nodes) // 2:]:
+        assert infraction_collector.infractions(0, node, 0) == True
```
cygnusv, line 154: * It would be good to define `RITUAL_ID = 0` and `MISSING_TRANSCRIPTS = 0` (probably part of its own enum) so the test is easier to read.
* Maybe a stylistic thing, but for boolean assertions, I'd prefer to omit `== True`, and conversely using `not` instead of `== False`. Of course, you can take it or leave it, it's not a strong opinion anyway.

contracts/contracts/coordination/InfractionCollector.sol
```diff
@@ -0,0 +1,78 @@
...
+            // Check if the infraction has already been reported
+            require(
+                !infractions[ritualId][stakingProviders[i]][InfractionType.MISSING_TRANSCRIPT],
+                "Infraction already reported"
+            );
+            Coordinator.Participant memory participant = coordinator.getParticipantFromProvider(
+                ritualId,
+                stakingProviders[i]
+            );
+            if (participant.transcript.length == 0) {
```
cygnusv, line 63: Rather than an `if` block here, we should convert this into a `require`, so that all nodes passed as parameters are infractors. Otherwise, we are (1) consuming more gas, and (2) allowing a scenario where a valid TX that includes good nodes can be source of confusion.

cygnusv (commented):
Some minor comments, overall looks very good! I'm fine if we decide to address my comments in a separate PR. BTW, there seems to be a conflict, so a rebase before merging might be in order

cygnusv (approved):

PR #269, opened 2024-05-16
Deploy updated TACoApplication contract that has updates for automated rewards

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Deployed updated version of TACoApplication as of 449ec475c433043d53bff2b55eaab91d9d71c60e. 
Updated TACoApplication deployment scripts for lynx.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/deployment/constructor_params/lynx/root.yml
+++ b/deployment/constructor_params/lynx/root.yml
@@ -8,6 +8,7 @@ artifacts:
 
 constants:
   IN_SECONDS_1_HOUR: 3600
+  IN_SECONDS_1_DAY: 86400
   IN_SECONDS_7_DAYS: 604800
   IN_SECONDS_60_DAYS: 5184000
   IN_SECONDS_91_DAYS: 7862400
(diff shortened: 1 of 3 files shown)


manumonti (approved):

deployment/constructor_params/lynx/root.yml
```diff
@@ -17,6 +18,8 @@ constants:
   FORTY_THOUSAND_TOKENS_IN_WEI_UNITS: 40000000000000000000000
   TEN_MILLION_TOKENS_IN_WEI_UNITS: 10000000000000000000000000
   YEAR_2025: 1735689600
+  PENALTY_DEFAULT: 1000 # 10%
+  PENALTY_INCREMENT: 500 # 5% increment
```
vzotova, line 22: My guess that for testnet we want those values to be bigger.

manumonti: > My guess that for testnet we want those values to be bigger.

How much would you choose for `default` and `increment`? @vzotova 

We were not sure about the numbers, so we decided (if I remember correctly) to have low quantities so we could play with a wider range of penalizations.

vzotova: considering testnet I'd do values like 50% default and 25% for increment, to make it more noticeable for anyone who uses testnet. for mainnet - yeah, smaller is better

cygnusv: I agree, but let's not redeploy because of that. In any case, we will measure reward amounts with scripts. We can change it later if needed.

cygnusv (approved):

PR #265, opened 2024-05-15
[WIP] Extract fees logic from Coordinator 

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
> High-level idea of the changes introduced in this PR. 
> List relevant API changes (if any), as well as related PRs and issues.

**Issues fixed/closed:**
Refs #264

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #270, opened 2024-05-17
[EPIC] New fee model architecture

### LINKED ISSUE #262: Cohort End of Service: Yellow and Red periods
Currently, cohorts have a duration defined in the Coordinator at initiation time. After this duration passes, cohorts are disbanded immediately. Although disbandment is not yet properly defined (see https://github.com/nucypher/nucypher/issues/3461), in the near future this will be an irreparable situation since it will involve destroying the secrets corresponding to the ritual.

We'd like to avoid a situation where cohort disbandment occurs due to adopters failing to renew their periodic payments. The proposed solution is to introduce two "grace periods" after the normal expiration date:
- 🟨 Yellow period: decryption requests are still serviced, but other management operations are blocked (e.g., authorizing encryptors)
- 🟥 Red period: cohort still exists, but no decryption request or management operation will be serviced. Once red period ends, cohort can disband and destroy their secrets.

Where and how these periods are defined is initially out of scope, as long as TACo nodes have a univocal way to retrieve them. My initial suggestion would be to include this as part of the `Subscription` contract (still TBD). Also note that current behavior is still  possible since it corresponds to `yellow = red = 0 seconds`.

### LINKED ISSUE #264: TACo Subscriptions
Initial design from @vzotova and yours truly:
* `Subscription` is a per-adopter contract that holds credit units (or simply `credits`).
* `Subscription` has an expiration date, but there's special periods after this date. See #262 
* Authorized contracts (like `Coordinator` and `AllowList` contracts) can "spend" credits from a `Subscription`.
* For the moment, `Subscription` contracts will be individual, with logic implemented in a case-by-case basis, but with the goal of generalizing a common factory contract down the road. The latter makes sense for the permissionless phase. For the moment, lets allow ourselves to experiment and offer what's best for each adopter.
* For the same reason as above, first subscription contracts will be upgradeable.
* Potentially, #263
* Subscription contract will specify rules for topping up (see #93 )
* Once service is charged (e.g., a ritual is initiated, an encryptor authoriation is realized, etc), fees may be sent directly to TG multisig. Alternatively, they can remain in the contract, but will be withdrawable by TG multisig.
* For the moment, let's handle refunds manually (i.e., involving the TG on a per-case basis).
* TG will be capable of "topping up" a Subscription contract with more credits if necessary (e.g., grants, testing, internal use, etc).
* `Subscriptions` can be seen as an evolution of the `BetaProgramInitiator` contract, which was only designed for ritual initiation.

Changes required in `Coordinator` and other contracts:
* #195 
* Coordinator doesn't handle fees directly anymore, just spends subscription credits.

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Changes fee model, now it's separate contract that must be approved in coordinator. 
Implements subscription for BqETH

**Issues fixed/closed:**
* Fixes #264 
* Fixes #262 

PRs:
- [x] #255 
- [x] #265 
- [x] #272 
- [x] #273 
- [x] #277 
- [x] #279 
- [x] #280 
- [x] #283 
- [x] #286 
- [x] #293 
- [x] #294
- [x] #295 
- [x] #296 
- [x] #297 
- [x] #298 
- [x] #299 

**Why it's needed:**

**Notes for reviewers:**
This PR is ready for tests and integration with nucypher repo, ManagedAllowList needs touch up in some of following PRs


contracts/contracts/coordination/GlobalAllowList.sol
```diff
@@ -4,74 +4,170 @@ pragma solidity ^0.8.0;
...
+     * @param _feeModel The address of the fee model contract
+     */
+    constructor(Coordinator _coordinator, IFeeModel _feeModel) {
+        require(
+            address(_coordinator) != address(0) && address(_feeModel) != address(0),
+            "Contracts cannot be zero addresses"
+        );
         require(_coordinator.numberOfRituals() >= 0, "Invalid coordinator");
         coordinator = _coordinator;
+        feeModel = _feeModel;
```
derekpierre, line 53: I'm slightly confused here. Is the fee model supposed to still be a constructor parameter? 

Instead, should the fee model for any ritual be obtained from the Coordinator instead of setting a static one in the constructor? i.e. there can be different fee models depending on the ritual id, and we have to look it up based on ritual id via the Coordinator.

Unless the "global allow list" is now an allow list for a specific ritual id, in which case perhaps a contract name change is warranted.

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -0,0 +1,261 @@
...
+            "Access controller for ritual must be approved"
+        );
+
+        if (activeRitualId != INACTIVE_RITUAL_ID) {
+            Coordinator.RitualState state = coordinator.getRitualState(activeRitualId);
+            require(
+                state == Coordinator.RitualState.DKG_INVALID ||
+                    state == Coordinator.RitualState.DKG_TIMEOUT ||
+                    state == Coordinator.RitualState.EXPIRED, // TODO check if it's ok
+                "Only failed rituals allowed to be reinitiated"
```
derekpierre, line 256: Since "expired" rituals aren't really "failed":
```suggestion
                "Only failed/expired rituals allowed to be reinitiated"
```

We could have used "non-active" but there is a specific `RitualState.ACTIVE` which gives "non-active" a different meaning than intended here.

contracts/contracts/coordination/subscription/Subscription.sol
```diff
@@ -0,0 +1,279 @@
...
+    function subscriptionFee() public pure returns (uint256) {
+        return 42 * 10 ** 20; // TODO
+    }
+
+    /**
+     * @notice Returns the base expiration duration
+     * @return The base expiration duration
+     */
+    function baseExpiration() public pure returns (uint256) {
+        return 52 weeks; // TODO
```
derekpierre, line 115: Do these "TODOs" still need to be revisited?

derekpierre (commented):
Looks like we will need some `nucypher` updates to accommodate the new architecture. Filed https://github.com/nucypher/nucypher/issues/3514.

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -591,13 +560,23 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+        require(msg.sender == ritual.initiator, "Only initiator can extend ritual");
+        require(getRitualState(ritual) == RitualState.ACTIVE, "Only active ritual can be extended");
+        ritual.endTimestamp += duration;
+        ritual.feeModel.processRitualExtending(
+            ritual.initiator,
+            ritualId,
+            ritual.participant.length,
+            duration
+        );
+        emit RitualExtended(ritualId, duration);
```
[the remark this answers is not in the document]
cygnusv: I think it's more informative to change duration for end timestamp

contracts/contracts/coordination/subscription/Subscription.sol
vzotova: I plan to remove Subscription contract when ManagedAllowList will be integrated with new Subscription

contracts/contracts/coordination/Coordinator.sol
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/coordination/GlobalAllowList.sol
vzotova: ✔️

derekpierre:
Just rebased this EPIC. Given the size of this PR - we should try to merge this soon. That being said, we should **not** merge it until v7.4.0 is released since merging it will cause `nucypher/nucypher` acceptance tests to fail because of contract changes. PR https://github.com/nucypher/nucypher/pull/3535 should remedy those issues moving forward once this PR is merged into `main`.

derekpierre (approved):
🎸

KPrasch (approved):

PR #276, opened 2024-06-14
Add `RewardsWithdrawn` event to `withdrawRewards` function in TACoApplication

### LINKED ISSUE #275: Adding `RewardsWithdrawn` event to TACoApp contract
Interface for Threshold applications includes an event `RewardsWithdrawn` intended to be emitted with the withdrawRewards is called (if it succeeds).

https://github.com/threshold-network/solidity-contracts/blob/8fcf42c6fcb50bb7730155bd5b99239b7b30382d/contracts/staking/IApplication.sol#L26

But in TACo application this event is not emitted. `RewardPaid` is emitted instead. 

https://github.com/nucypher/nucypher-contracts/blob/07ce241f1186e46342fc74a213ab81d5a791862f/contracts/contracts/TACoApplication.sol#L559

So, if we want to fit with this interface, a `RewardsWithdrawn` event should be emitted.

Considering that  IApplication's `RewardsWithdrawn` has two parameters and TACoApplication's `RewardPaid` has four parameters, I see two feasible solutions:

1. `withdrawRewards` function in `TACoApplication` contract emits the IApplication's `RewardsWithdrawn` event in addition to the currently emitted `RewardPaid` event.

2. Modify IApplication's `RewardsWithdrawn` event so the parameters are the same as `RewardPaid` (`sender | stakingProvider | beneficiary | reward` instead of `stakingProvider | reward`). This solution is probably less realistic since this will need to update deployed contracts like tBTC and RandomBeacon:

https://github.com/keep-network/keep-core/blob/7b46b2179e525e848ced5fe67445c109da8f3dfb/solidity/ecdsa/contracts/WalletRegistry.sol#L364

Related issue: https://github.com/threshold-network/solidity-contracts/issues/156

manumonti:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
This will add `RewardsWithdrawn` to be emitted when staking rewards are withdrawn.

**Issues fixed/closed:**
Fixes #275 

**Why it's needed:**
This event is included in `IApplication` interface, so it's intended to be emitted by all the Threshold applications when the function `withdrawRewards()` is called.

https://github.com/threshold-network/solidity-contracts/blob/8fcf42c6fcb50bb7730155bd5b99239b7b30382d/contracts/staking/IApplication.sol#L26


vzotova (approved):

manumonti:
TBD: tests

vzotova (approved):

derekpierre (approved):
🎸

PR #283, opened 2024-07-03
BqETHSubscription: base fee rate increase

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/coordination/subscription/BqETHSubscription.sol
+++ b/contracts/contracts/coordination/subscription/BqETHSubscription.sol
@@ -22,17 +22,16 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
     }
 
     uint32 public constant INACTIVE_RITUAL_ID = type(uint32).max;
-    uint192 public constant INCREASE_BASE = 10000;
+    uint256 public constant INCREASE_BASE = 10000;
 
     GlobalAllowList public immutable accessController;
     IERC20 public immutable feeToken;
     address public immutable adopter;
 
     uint256 public immutable initialBaseFeeRate;
+    uint256 public immutable baseFeeRateIncrease;
     uint256 public immutable encryptorFeeRate;
     uint256 public immutable maxNodes;
-
-    uint256 public immutable baseFeeRateIncrease;
 
     uint32 public activeRitualId;
     mapping(uint256 periodNumber => Billing billing) public billingInfo;
@@ -115,6 +114,10 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
             address(_accessController) != address(0),
             "Access controller cannot be the zero address"
         );
+        require(
+            _baseFeeRateIncrease < INCREASE_BASE,
+            "Base fee rate increase must be fraction of INCREASE_BASE"
+        );
         feeToken = _feeToken;
         adopter = _adopter;
         initialBaseFeeRate = _initialBaseFeeRate;
@@ -156,8 +159,9 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
 
     function baseFees(uint256 periodNumber) public view returns (uint256) {
         uint256 baseFeeRate = initialBaseFeeRate *
-            (INCREASE_BASE + baseFeeRateIncrease * periodNumber);
-        return (baseFeeRate * subscriptionPeriodDuration * maxNodes) / INCREASE_BASE;
+            (INCREASE_BASE + baseFeeRateIncrease) ** periodNumber;
+        return
+            (baseFeeRate * subscriptionPeriodDuration * maxNodes) / (INCREASE_BASE ** periodNumber);
     }
 
     function encryptorFees(uint128 encryptorSlots, uint32 duration) public view returns (uint256) {
--- a/contracts/contracts/coordination/subscription/BqETHSubscription.sol
+++ b/contracts/contracts/coordination/subscription/BqETHSubscription.sol
@@ -157,6 +157,7 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
         return baseFees(currentPeriodNumber);
     }
 
+    /// @dev optential overflow after 15-16 periods
     function baseFees(uint256 periodNumber) public view returns (uint256) {
         uint256 baseFeeRate = initialBaseFeeRate *
             (INCREASE_BASE + baseFeeRateIncrease) ** periodNumber;
--- a/contracts/contracts/coordination/subscription/BqETHSubscription.sol
+++ b/contracts/contracts/coordination/subscription/BqETHSubscription.sol
@@ -21,17 +21,19 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
         uint128 encryptorSlots; // pre-paid encryptor slots for the billing period
     }
 
-    IERC20 public immutable feeToken;
-
     uint32 public constant INACTIVE_RITUAL_ID = type(uint32).max;
+    uint192 public constant INCREASE_BASE = 10000;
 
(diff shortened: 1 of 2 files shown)


contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -145,7 +150,14 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
...
     function baseFees() public view returns (uint256) {
-        return baseFeeRate * subscriptionPeriodDuration * maxNodes;
+        uint256 currentPeriodNumber = getCurrentPeriodNumber();
+        return baseFees(currentPeriodNumber);
+    }
+
+    function baseFees(uint256 periodNumber) public view returns (uint256) {
+        uint256 baseFeeRate = initialBaseFeeRate *
+            (INCREASE_BASE + baseFeeRateIncrease * periodNumber);
+        return (baseFeeRate * subscriptionPeriodDuration * maxNodes) / INCREASE_BASE;
```
cygnusv, line 160: This implementation doesn't factor in compounding interest (i.e., rate increases should be applied over the previous period base rate, not over the initial base rate). We could do something like this:
```suggestion
            (INCREASE_BASE + baseFeeRateIncrease) ** periodNumber;
        return (baseFeeRate * subscriptionPeriodDuration * maxNodes) / (INCREASE_BASE ** periodNumber);
```

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -21,17 +21,19 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
         uint128 encryptorSlots; // pre-paid encryptor slots for the billing period
     }
 
-    IERC20 public immutable feeToken;
-
     uint32 public constant INACTIVE_RITUAL_ID = type(uint32).max;
+    uint192 public constant INCREASE_BASE = 10000;
```
cygnusv, line 25: Any particular reason this is uint192 and not uint256?

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -21,17 +21,19 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
...
+    IERC20 public immutable feeToken;
     address public immutable adopter;
 
-    uint256 public immutable baseFeeRate;
+    uint256 public immutable initialBaseFeeRate;
     uint256 public immutable encryptorFeeRate;
     uint256 public immutable maxNodes;
 
-    GlobalAllowList public immutable accessController;
+    uint256 public immutable baseFeeRateIncrease;
```
cygnusv, line 35: just nitpicking, but can we place this a few lines above, together with the other fee rate stuff?

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -113,7 +117,8 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
         );
         feeToken = _feeToken;
         adopter = _adopter;
-        baseFeeRate = _baseFeeRate;
+        initialBaseFeeRate = _initialBaseFeeRate;
+        baseFeeRateIncrease = _baseFeeRateIncrease;
```
cygnusv, line 121: We need to validate that `_baseFeeRateIncrease` is a fraction of `INCREASE_BASE`.

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: right! I'll change it

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: my bad, when copied - just missed that type is not 256, I'll change it

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: actually I thought that nothing prevents to have increase more than 100% so basically could be any number, but probably it's not great idea, yeah I can restrict it

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: ✔️

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: ✔️

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: ✔️

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: ✔️

derekpierre (approved):
🎸

tests/test_bqeth_subscription.py
```diff
@@ -35,7 +36,16 @@
 YELLOW_PERIOD = ONE_DAY
```
arjunhassard, line 36: Just want to check that the base fee rate will increase (e.g. on a yearly basis) regardless of whether the ritual ends up falling into the yellow or red grace periods, and regardless of when they renew.

arjunhassard (approved):
Nice, super quick turnaround. Can this be included in the BqETH initialization in mid-July or will we have to upgrade later?

cygnusv:
> Can this be included in the BqETH initialization in mid-July

@arjunhassard Absolutely

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -145,7 +153,15 @@ contract BqETHSubscription is EncryptorSlotsSubscription, Initializable, Ownable
...
 
     function baseFees() public view returns (uint256) {
-        return baseFeeRate * subscriptionPeriodDuration * maxNodes;
+        uint256 currentPeriodNumber = getCurrentPeriodNumber();
+        return baseFees(currentPeriodNumber);
+    }
+
+    function baseFees(uint256 periodNumber) public view returns (uint256) {
+        uint256 baseFeeRate = initialBaseFeeRate *
+            (INCREASE_BASE + baseFeeRateIncrease) ** periodNumber;
```
cygnusv, line 162: Maybe we should add a note here, but this can overflow after 15 or 16 periods. I don't think it's a major concern, but at least a dev note would be OK. There's also iterative implementations that would remove the possibility of overflow, but they will increase code complexity, and it's probably not worth it.

tests/test_bqeth_subscription.py
cygnusv: Yes, I think it's the case because base fee rate increase depends exclusively on which period number we are at the moment, and that only depends on the period duration and the initial subscription timestamp. In any case, let's wait for confirmation from Vicky

cygnusv (approved):

tests/test_bqeth_subscription.py
vzotova: @cygnusv 100% correct

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: ✔️

PR #289, opened 2024-07-09
Bqeth testing and usage script

theref:
Closes https://github.com/orgs/nucypher/projects/55/views/1?pane=issue&itemId=69683595


scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,171 @@
...
+    ) = setup_contracts()
+
+    SUBSCRIPTION_PERIOD = subscription_contract.functions.subscriptionPeriodDuration().call()
+    YELLOW_PERIOD = subscription_contract.functions.yellowPeriodDuration().call()
+    RED_PERIOD = subscription_contract.functions.redPeriodDuration().call()
+
+    base_fees = subscription_contract.functions.baseFees(0).call()
+    encryptor_fees = subscription_contract.functions.encryptorFees(
+        MAX_NODES, SUBSCRIPTION_PERIOD
+    ).call()
```
cygnusv, line 164: It'd be great to have these two calls as separate functions

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,171 @@
...
+    registry = json.load(f)
+
+SUBSCRIPTION_CONTRACT_ADDRESS = registry[CHAIN]["BqETHSubscription"]["address"]
+SUBSCRIPTION_CONTRACT_ABI = registry[CHAIN]["BqETHSubscription"]["abi"]
+ERC20_CONTRACT_ADDRESS = registry[CHAIN][NUCYPHER_DOMAIN.title() + "RitualToken"]["address"]
+ERC20_CONTRACT_ABI = registry[CHAIN][NUCYPHER_DOMAIN.title() + "RitualToken"]["abi"]
+COORDINATOR_CONTRACT_ADDRESS = registry[CHAIN]["Coordinator"]["address"]
+COORDINATOR_CONTRACT_ABI = registry[CHAIN]["Coordinator"]["abi"]
+GLOBAL_ALLOW_LIST_CONTRACT_ADDRESS = registry[CHAIN]["GlobalAllowList"]["address"]
+GLOBAL_ALLOW_LIST_CONTRACT_ABI = registry[CHAIN]["GlobalAllowList"]["abi"]
```
KPrasch, line 32: I see that you kept this as a pure web3.py script, but if you want to introduce ape functionality and custom utilities, there's a utility in `deployments.registry` that will clean this up a bit.

```
from deployment.registry import contracts_from_registry

contracts = contracts_from_registry('deployment/artifacts/lynx.json', 80002)
...
coordinator = contracts["Coordinator"]
...
coordinator.address
```

KPrasch: 👍🏻 What's the reason for that?

theref: when David and I chatted about it we agreed that keeping it as vanilla as possible was a good idea. It's to act as a demo for adopters, we don't want to force them into using `ape`

manumonti (approved):
LGTM! 👌

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+# TODO: make the provider a parameter on the CLI
+
+with open(f"deployment/artifacts/{NUCYPHER_DOMAIN}.json", "r") as f:
+    registry = json.load(f)
+
+# SUBSCRIPTION_CONTRACT_ADDRESS = registry[CHAIN]["BqETHSubscription"]["address"]
+SUBSCRIPTION_CONTRACT_ADDRESS = registry[CHAIN]["FreeFeeModel"]["address"]
+SUBSCRIPTION_CONTRACT_ABI = registry[CHAIN]["FreeFeeModel"]["abi"]
+# ERC20_CONTRACT_ADDRESS = registry[CHAIN][NUCYPHER_DOMAIN.title() + "RitualToken"]["address"]
+# ERC20_CONTRACT_ABI = registry[CHAIN][NUCYPHER_DOMAIN.title() + "RitualToken"]["abi"]
```
KPrasch, line 68: Do you intend to keep these lines commented out?

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+
+import click
+import requests
+from dotenv import load_dotenv
+from web3 import Web3
+from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
+
+# Load environment variables
+load_dotenv()
+PROVIDER_URL = os.environ.get("PROVIDER_URL")
```
KPrasch, line 55: This can be a CLI parameter instead of an env-var.  This may alleviate some potential issues when changing providers between mainnet and testnet while testing.  Also, if you want to keep the env var functionality, click provides some out-of-the-box functionality for this:

https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables
https://click.palletsprojects.com/en/8.1.x/options/#dynamic-defaults-for-prompts

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+from collections import namedtuple
+
+import click
+import requests
+from dotenv import load_dotenv
+from web3 import Web3
+from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
+
+# Load environment variables
+load_dotenv()
```
KPrasch, line 54: Note from the `dotenv` documentation

> By default, load_dotenv doesn't override existing environment variables.

This means that in a single terminal session, this script can only load env variables once.  If you think it's appropriate/safe you may consider using `load_dotenv(override=True)`.

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+    #     + context.subscription_contract.functions.yellowPeriodDuration().call()
+    #     + context.subscription_contract.functions.redPeriodDuration().call()
+    # )
+    # if start_of_subscription > 0:
+    #     click.echo(
+    #         "Subscription has already started. Subtracting the elapsed time from the duration."
+    #     )
+    #     now = context.w3.eth.get_block("latest")["timestamp"]
+    #     elapsed = now - start_of_subscription + 100
+    #     duration -= elapsed
```
KPrasch, line 210: Why is this block commented out?

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+    for i in range(num_rituals):
+        click.echo(f"Initiating ritual {i + 1} of {num_rituals}.")
+        nodes = sorted(
+            [
+                u["checksum_address"]
+                for u in requests.get(porter_request).json()[
+                    "result"
+                ]["ursulas"]
+            ], key=lambda x: x.lower()
+        )
```
KPrasch, line 225: This list comprehension is a bit unreadable, perhaps split this up into a few lines with variables instead.

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@cli.command()
+@click.option("--encryptor-slots", default=2, help="Number of encryptor slots to pay for.")
+@setup_context
+def pay_subscription(context, encryptor_slots):
+    """Pay for a new subscription period and initial encryptor slots."""
+    base_fees = context.subscription_contract.functions.baseFees(0).call()
+    encryptor_fees = context.subscription_contract.functions.encryptorFees(
+        MAX_NODES, context.subscription_contract.functions.subscriptionPeriodDuration().call()
```
KPrasch, line 155: What is the use of `MAX_NODES` here?  Looking at the contract (https://github.com/nucypher/nucypher-contracts/blob/epic-subscription/contracts/contracts/coordination/subscription/BqETHSubscription.sol#L195). It appears that this is not a maximum but instead the number of initial encryptor slots to pay for?

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+
+@cli.command()
+@click.option("--extra-slots", default=1, help="Number of additional encryptor slots to pay for.")
+@setup_context
+def pay_slots(context, extra_slots):
+    """Pay for additional encryptor slots."""
+    click.echo(f"Paying for {extra_slots} new encryptor slots.")
+    tx_hash = context.subscription_contract.functions.payForEncryptorSlots(extra_slots).transact(
+        {"from": context.account.address}
+    )
```
KPrasch, line 181: 1. Does this function also imply the transfer of erc20 fees to the subscription contract?  
2. If so - we may need to include erc20 approval here also?  
3. Perhaps this is related to how you handle `MAX_NODES` above as a means of "pre-approving" an ERC20 amount?

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+@cli.command()
+@click.option("--num-nodes", default=2, help="Number of nodes to use for the ritual.")
+@click.option("--duration", help="Duration in seconds for the ritual.", type=int)
+@click.option("--random-seed", help="Random seed integer for sampling.")
+@click.option("--num-rituals", default=1, help="Number of rituals to initiate.", type=int)
+@setup_context
+def initiate_ritual(context, num_nodes, duration, random_seed, num_rituals):
+    """Initiate a ritual."""
+    if NUCYPHER_DOMAIN == "mainnet":
+        porter_request = f"https://porter.nucypher.community/bucket_sampling?quantity={num_nodes}&random_seed={random_seed}"
```
KPrasch, line 194: Instead of URL-encoding in a string template, requests does have the ability to accept a dictionary of `params`.

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+    #     click.echo(
+    #         "Subscription has already started. Subtracting the elapsed time from the duration."
+    #     )
+    #     now = context.w3.eth.get_block("latest")["timestamp"]
+    #     elapsed = now - start_of_subscription + 100
+    #     duration -= elapsed
+
+    click.echo(f"Account address that will sign the transaction: {context.account.address}")
+    click.echo(f"Initiating ritual with {num_nodes} providers for {duration} seconds.")
+    click.echo(f"Endpoint used: {PROVIDER_URL}")
```
KPrasch, line 214: These can be combined into a single expression.

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+                for u in requests.get(porter_request).json()[
+                    "result"
+                ]["ursulas"]
+            ], key=lambda x: x.lower()
+        )
+
+        tx_hash = context.coordinator_contract.functions.initiateRitual(
+            context.subscription_contract.address,
+            nodes,
+            context.account.address, # TODO: parametrize this
```
KPrasch, line 230: This is the authority address of the ritual.  For production use-cases this will not be the same as the initiator so we need to parameterize this.

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+    click.echo(f"Transaction hash: {tx_hash.hex()}")
+
+
+@cli.command()
+@click.option("--num-nodes", default=2, help="Number of nodes to use for the ritual.")
+@click.option("--duration", help="Duration in seconds for the ritual.", type=int)
+@click.option("--random-seed", help="Random seed integer for sampling.")
+@click.option("--num-rituals", default=1, help="Number of rituals to initiate.", type=int)
+@setup_context
+def initiate_ritual(context, num_nodes, duration, random_seed, num_rituals):
```
KPrasch, line 191: This function has a sufficiently different scope than subscription payment/management and will be executed by a different permissioned role. IMO this command can be removed from this PR entirely.  In fact, there is already a ritual initiation script in this same directory.

scripts/bqeth_subscription.py
theref: i think some commit have been pushed that shouldn't after the script was used locally for some testing

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+        "coordinator_contract",
+        "global_allow_list_contract",
+    ],
+)
+
+
+def setup_connections():
+    """Set up Web3 connections."""
+    w3 = Web3(Web3.HTTPProvider(PROVIDER_URL))
+    w3.middleware_onion.inject(geth_poa_middleware, layer=0)
```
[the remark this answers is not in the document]
theref: correct

scripts/bqeth_subscription.py
vzotova: yep, you are correct, or it could be number of additional encryptor slots

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,264 @@
...
+from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
+
+# Load environment variables
+load_dotenv()
+PROVIDER_URL = os.environ.get("PROVIDER_URL")
+PRIVATE_KEY = os.environ.get("PRIVATE_KEY")
+
+MAX_NODES = 10
+NUCYPHER_DOMAIN = "mainnet"
+CHAIN = "137"
```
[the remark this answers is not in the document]
theref: fixed in aef0976

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,267 @@
...
+    "Context",
+    [
+        "w3",
+        "account",
+        "erc20_contract",
+        "subscription_contract",
+        "coordinator_contract",
+        "global_allow_list_contract",
+    ],
+)
```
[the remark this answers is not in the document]
theref: fixed in aef0976

scripts/bqeth_subscription.py
theref: fixed in aef0976

scripts/bqeth_subscription.py
theref: fixed in aef0976

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,266 @@
...
+                f"{context.constants['PORTER_ENDPOINT']}/get_ursulas?quantity={num_nodes}"
+            ).json()["result"]["ursulas"]
+        ]
+    ))
+    start_of_subscription = context.subscription_contract.functions.startOfSubscription().call()
+    duration = (
+        context.subscription_contract.functions.subscriptionPeriodDuration().call()
+        + context.subscription_contract.functions.yellowPeriodDuration().call()
+        + context.subscription_contract.functions.redPeriodDuration().call()
+    )
```
KPrasch, line 216: Ah - This is very interesting functionality!  Just making sure I understand correctly it looks like you are correlating a ritual's duration with the specific temporal rules of a subscription?

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,266 @@
...
+    if start_of_subscription > 0:
```
KPrasch, line 217: Is `startOfSubscription` a fixed timestamp or a calculated period since the "start"? cc @vzotova

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,266 @@
...
+import requests
+from dotenv import load_dotenv
+from web3 import Web3
+from web3.contract import Contract
+from web3.middleware import construct_sign_and_send_raw_middleware, geth_poa_middleware
+from eth_account.signers.local import LocalAccount
+
+
+def build_constants(domain, provider_url, private_key):
+    DOMAIN_TO_CHAIN = {"lynx": "80002", "tapir": "80002", "mainnet": "137"}
```
KPrasch, line 57: There are string constants for these domain strings in `deployments/constants.py`

```suggestion
    DOMAIN_TO_CHAIN = {LYNX: "80002", TAPIR: "80002", MAINNET: "137"}
```

scripts/bqeth_subscription.py
```diff
@@ -0,0 +1,266 @@
...
+        registry = json.load(f)
+
+    return {
+        "PROVIDER_URL": provider_url,
+        "PRIVATE_KEY": private_key,
+        "PORTER_ENDPOINT": PORTER_ENDPOINT,
+        "SUBSCRIPTION_CONTRACT_ADDRESS": registry[chain]["BqETHSubscription"]["address"],
+        "SUBSCRIPTION_CONTRACT_ABI": registry[chain]["BqETHSubscription"]["abi"],
+        "ERC20_CONTRACT_ADDRESS": registry[chain][f"{domain.title()}RitualToken"]["address"],
+        "ERC20_CONTRACT_ABI": registry[chain][f"{domain.title()}RitualToken"]["abi"],
```
KPrasch, line 71: Is this script intended only for testnets?

theref:
@KPrasch do we still need this with #299 ?

KPrasch:
> @KPrasch do we still need this with #299 ?

Do you mean this entire PR? This script broke the ice for testnet fee model deployments but I'm in favor of closing this PR as long as you feel that all of the useful bits have been included elsewhere.  Also, I want to acknowledge that #299 is based off your initial work here (I made sure to include you as a commit author).  

That being said (almost) all of the functionality implemented in this PR has been implemented in #299 and #297.  The main difference is that this PR has fewer dependencies (#299 heavily depends on eth-ape abstractions).  Personally, I do not view that as a major issue.  There are a few interesting bits of logic here, particularly with regard to ritual duration calculation that I'd like to understand better so we may possibly include the same changes in #297.

scripts/bqeth_subscription.py
vzotova: it's fixed after initiating ritual

scripts/bqeth_subscription.py
KPrasch: @theref Can you make a commit into PR #297 to include these calculations?

KPrasch: #297 is merged now but it will still be good to include these changes in `initiate_ritual.py`  @theref Do you want to take care of that?

KPrasch:
This branch was closed by automation after deleting the epic-subscription branch.  I suggest making a new PR into `main` for any additional follow-ups.

PR #296, opened 2024-07-29
Updated TACoApplication deployment

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


deployment/artifacts/mainnet.json
```diff
@@ -149,7 +149,7 @@
             "deployer": "0xFfFd7092685bDeeBD121D1A0FEA3c349114Cce50"
         },
         "TACoApplication": {
-            "address": "0x347CC7ede7e5517bD47D20620B2CF1b406edcF07",
+            "address": "0x98F643B32b5A3513929Dfd3F2E3e39F7376659DB",
```
derekpierre, line 152: Was this deployed as a proxy upgrade? I thought TACoApplication was proxied, in which case the address shouldn't change.

derekpierre: ^ @cygnusv

cygnusv:
```suggestion
            "address": "0x347CC7ede7e5517bD47D20620B2CF1b406edcF07",
```

derekpierre: I'm wondering if there may be an issue in the deployment script itself if this needed to be manually reverted. I believe for proxy deployment upgrades the `Deployer` typically handles not updating the address but updating the ABI in the resulting contract registry.

I'll try to take a deeper look at the deployment script.

(cc @KPrasch )

KPrasch (approved):

derekpierre (changes requested):
Just putting this marker for now until we take a deeper look at the deployment script.

deployment/artifacts/mainnet.json
derekpierre: I think I get it now.

On mainnet we can't actually upgrade the proxied contracts ourselves; that can only be done by the Council Multisig. As a result we only deploy implementation contracts, and then have the Council subsequently upgrade the proxy to point to the implementation contract. However, our deployment scripts are not accustomed to the proxy upgrade process being split into two separate operations.

As a result, when we deploy the implementation contract on the `mainnet,` the contract registry address gets changed even for a proxy contract. 

For now we have to be careful not to update the addresses in the contract registry, and maybe in the future we can come up with an option to not have the registry update the address of the contract.

In the context of this PR we are good now that the address change has been reverted to the original address.

That being said, we should not merge this PR into `main` until the Council has upgraded the TACoApplication proxy contract (unless that has already happened).

derekpierre (approved):
🎸

derekpierre:
Council already upgraded the proxy contract - https://etherscan.io/tx/0xc3adf42427bbf4f25026da756a02e47317e11bd6ae6cebe8303b38b72f660370

PR #299, opened 2024-08-05
Ape-based generic subscription and authorization management CLI

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
Introduces subscription-generic CLI that can be used for long-term management of multiple subscription contracts and authorizations for multiple adopters.  In contrast to PR #289 this PR implements both local and ape utilities.  As a result, it's necessary to install all the dependencies.

**Notes for reviewers:**
- Heavily inspired/adapted original work committed by @theref in PR #289
- Commits are co-authored by @theref and @derekpierre

### DIFF
--- a/scripts/manage_subscription.py
+++ b/scripts/manage_subscription.py
@@ -161,7 +161,7 @@ def remove_encryptors(account, domain, ritual_id, access_controller, encryptors)
     )
     click.echo(
         f"Removing {len(encryptors)} "
-        f"encryptors to the {access_controller} "
+        f"encryptors from the {access_controller} "
         f"for ritual {ritual_id}."
     )
     transactor.transact(
--- a/scripts/manage_subscription.py
+++ b/scripts/manage_subscription.py
@@ -1,0 +1,174 @@
+import click
+from ape import Contract
+from ape.cli import account_option, ConnectedProviderCommand
+
+from deployment import registry
+from deployment.options import (
+    subscription_contract_option,
+    domain_option,
+    access_controller_option,
+    ritual_id_option,
+    encryptor_slots_option,
+    encryptors_option,
+)
+from deployment.params import Transactor
+from deployment.utils import check_plugins
+
+
+def _erc20_approve(
+        amount: int,
+        erc20: Contract,
+        receiver: Contract,
+        transactor: Transactor
+) -> None:
+    """Approve an ERC20 transfer."""
+    click.echo(
+        f"Approving transfer of {amount} {erc20.contract_type.name} "
+        f"to {receiver.contract_type.name}."
+    )
+    transactor.transact(
+        erc20.approve,
+        receiver.address,
+        amount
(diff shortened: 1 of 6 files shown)


theref (approved):

manumonti (approved):
This is going to be very useful 👌

vzotova (approved):

scripts/manage_subscription.py
```diff
@@ -0,0 +1,174 @@
...
+@account_option()
+@domain_option
+@subscription_contract_option
+@encryptor_slots_option
+@click.option(
+    "--period",
+    default=0,
+    help="Subscription billing period number to pay for.",
+)
+def pay_subscription(account, domain, subscription_contract, encryptor_slots, period):
```
derekpierre, line 62: If you wanted these commands to not use dashes in the CLI you could use the name option in the command() directive. For example, for this function you could do:
```python
@cli.command(name=paySubscription, cls=ConnectedProviderCommand)
...
```
Then the command would not use any dashes.

You can do the same for the `click.group()` i.e. `click.group(name="manageSubscription") for the `manage_subscription` CLI command group which derives its name from the file by default.

Up to you though - just giving options.

scripts/manage_subscription.py
```diff
@@ -0,0 +1,174 @@
...
+def remove_encryptors(account, domain, ritual_id, access_controller, encryptors):
+    """Deauthorize encryptors from the access control contract for a ritual."""
+    transactor = Transactor(account=account)
+    access_controller = registry.get_contract(
+        contract_name=access_controller,
+        domain=domain
+    )
+    click.echo(
+        f"Removing {len(encryptors)} "
+        f"encryptors to the {access_controller} "
```
derekpierre, line 164:
```suggestion
        f"encryptors from the {access_controller} "
```

scripts/manage_subscription.py
KPrasch: I gave this a try for the `click.group` but I did not have any success 🤷🏻 

```python
@click.group(name="manage-subscription")
def cli():
    """Subscription Management CLI"""
```

```python
(nucypher-contracts) k@Kierans-MacBook-Pro nucypher-contracts % ape run manage-subscription --help
Usage: ape run [OPTIONS] COMMAND [ARGS]...
Try 'ape run -h' for help.

Error: No such command 'manage-subscription'. Did you mean manage_subscription?
```

derekpierre: Interesting. I guess the cli group name doesn't matter to ape when doing `ape run ...`.

derekpierre: The "name" property for click.command should still work though (if desired).

KPrasch: For sure, it works.  Personally I prefer the default naming behavior for CLI commands.

PR #306, opened 2024-08-15
Add duration calculation for subscription based rituals

theref:
If anyone wants to test it here is the command:

`ape run initiate_ritual --network polygon:amoy:infura -d lynx -c GlobalAllowList -f BqETHSubscription -a 0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600 --num-nodes 3`

### DIFF
--- a/scripts/initiate_ritual.py
+++ b/scripts/initiate_ritual.py
@@ -120,10 +120,13 @@ def cli(
             + fee_model_contract.redPeriodDuration()
         )
         if start_of_subscription > 0:
+            end_of_subscription = fee_model_contract.getEndOfSubscription()
+            now = chain.blocks.head.timestamp
+            if now > end_of_subscription:
+                raise ValueError("Subscription has already ended.")
             click.echo(
             "Subscription has already started. Subtracting the elapsed time from the duration."
         )
-            now = chain.blocks.head.timestamp
             elapsed = now - start_of_subscription + 100
             duration -= elapsed
 


derekpierre:
I can run it locally but get the following error:
```
    result = sorted(data["result"]["ursulas"], key=lambda x: x.lower())
                    ~~~~^^^^^^^^^^
KeyError: 'result'
```

Porter is returning a 400 because the `duration` calculated is negative:
```
{'error': 'duration must be a non-negative integer.', 'version': '3.7.0'}
```

The duration value used as a parameter is:
```
'duration': -2804636
```
Therefore, no results are returned.

As a side note, the `sample_nodes(...)` utility function could probably handle any errors from porter sampling better i.e. check the response status code before assuming that the call was successful.

theref:
@derekpierre thank you! Ape seems happier today (i think there's something odd with `ape-etherscan` api urls for polygon testnets, I'm looking into it...) and I got the same error - it's happening because the subscription has already ended. I've added a check for this in https://github.com/nucypher/nucypher-contracts/pull/306/commits/2619079d5ec0b8916d856c9b70ee5babbb19ec51

vzotova (approved):

scripts/initiate_ritual.py
```diff
@@ -105,6 +106,31 @@ def cli(
...
+            + fee_model_contract.redPeriodDuration()
+        )
+        if start_of_subscription > 0:
+            end_of_subscription = fee_model_contract.getEndOfSubscription()
+            now = chain.blocks.head.timestamp
+            if now > end_of_subscription:
+                raise ValueError("Subscription has already ended.")
+            click.echo(
+            "Subscription has already started. Subtracting the elapsed time from the duration."
+        )
```
cygnusv, line 129: Are these 2 lines properly linted?

cygnusv (approved):

PR #307, opened 2024-08-15
Coordinator: moves rituals from array to mapping

vzotova:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -131,6 +131,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         __AccessControlDefaultAdminRules_init(0, _admin);
     }
 
+    /// @dev use `upgradeAndCall` for upgrading together with re-initialization
     function initializeNumberOfRituals() external reinitializer(2) {
         if (numberOfRituals == 0) {
             numberOfRituals = ritualsStub.length;
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -93,7 +93,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     ITACoChildApplication public immutable application;
     uint96 private immutable minAuthorization; // TODO use child app for checking eligibility
 
-    Ritual[] public rituals;
+    Ritual[] private ritualsStub; // former rituals
     uint32 public timeout;
     uint16 public maxDkgSize;
     bool private stub1; // former isInitiationPublic
@@ -106,10 +106,13 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     mapping(address => ParticipantKey[]) internal participantKeysHistory;
     mapping(bytes32 => uint32) internal ritualPublicKeyRegistry;
     mapping(IFeeModel => bool) public feeModelsRegistry;
+
+    mapping(uint256 index => Ritual ritual) internal _rituals;
+    uint256 public numberOfRituals;
     // Note: Adjust the __preSentinelGap size if more contract variables are added
 
     // Storage area for sentinel values
-    uint256[19] internal __preSentinelGap;
+    uint256[17] internal __preSentinelGap;
     Participant internal __sentinelParticipant;
     uint256[20] internal __postSentinelGap;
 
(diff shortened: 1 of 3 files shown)


tests/test_coordinator.py
```diff
@@ -564,3 +565,71 @@ def test_post_aggregation_fails(
...
+    assert end - init == DURATION
+    total_transcripts, total_aggregations = ritual_struct[3], ritual_struct[4]
+    assert total_transcripts == total_aggregations == 0
+    assert ritual_struct[5] == initiator
+    assert ritual_struct[6] == len(nodes)
+    assert ritual_struct[7] == 1 + len(nodes) // 2  # threshold
+    assert not ritual_struct[8]  # aggregationMismatch
+    assert ritual_struct[9] == global_allow_list.address  # accessController
+    assert ritual_struct[10] == (b"\x00" * 32, b"\x00" * 16)  # publicKey
+    assert not ritual_struct[11]  # aggregatedTranscript
```
derekpierre, line 607: If I'm understanding this test correctly, you are creating and checking 3 rituals:
1. first one mimics ritual that was created before any concept of fee model - fee model is 0x...0
2. second mimics ritual that was created with a fee model address but was initiated with the storage before your current storage change fix
3. third mimics ritual that was created with your current storage change fix

For `v7.5.x` (so unreleased code for next release) we use index 12 for fee model - https://github.com/nucypher/nucypher/commit/2c2cb5dc05f0650d5f95bf5a4042dd262849c06f#diff-911a370c18d0d0e3ea62099fc5ff6daa16915da19cf5399f24e404b7d8d6d168R597. 
- Is that still fine or will that be problematic?
- If not problematic, should we also check the values in the struct at index `12` (fee model) in this test?

vzotova: for client that supports `new fee model architecture` - everything should stay the same, so yeah feeModel is 12th in structure.
all add test in check

vzotova: added this check
interesting that `ritual_struct[12]` is participants array and not fee model
I'm not sure if it's ape behavior or abi enhanced for nested structs comparing to what was before, anyway could you check if 7.5.x would work with this update, please

derekpierre: > interesting that ritual_struct[12] is participants array and not fee model

Indeed. This was how we realized we needed to get participants for a ritual in a separate call.

> could you check if 7.5.x would work with this update, please

Ran the acceptance tests for `v7.5.x` with the following update to the ape-config.yaml:
```diff
diff --git a/tests/acceptance/ape-config.yaml b/tests/acceptance/ape-config.yaml
index b44995ae6..3d3d34ce2 100644
--- a/tests/acceptance/ape-config.yaml
+++ b/tests/acceptance/ape-config.yaml
@@ -5,8 +5,8 @@ plugins:

 dependencies:
   - name: nucypher-contracts
-    github: nucypher/nucypher-contracts
-    ref: main
+    github: vzotova/nucypher-contracts
+    ref: coordinator-layout
     config_override:
       solidity:
         version: 0.8.23
```

and the acceptance tests all pass without any changes 🎉 .

derekpierre (approved):
🎸

theref (approved):

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -128,35 +131,57 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         __AccessControlDefaultAdminRules_init(0, _admin);
     }
 
+    function initializeNumberOfRituals() external reinitializer(2) {
```
cygnusv, line 134: Should we include something to make sure that this function is called before the contract is used?

cygnusv (approved):

contracts/contracts/coordination/Coordinator.sol
vzotova: I thought about that but honestly - it will only increase gas for checking and we can upgrade together with reinitialize (should be possible)

cygnusv: Maybe let's add a dev comment reminding to use `upgradeAndCall` do it.

vzotova: ✔️

derekpierre:
Do we need a follow-up issue with a rollout plan for these changes? See #309
(@cygnusv , @vzotova, @arjunhassard  )

PR #312, opened 2024-08-21
Upgrade lynx Coordinator

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -138,8 +138,43 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         }
     }
 
-    function rituals(uint32 ritualId) public view returns (Ritual memory) {
-        return storageRitual(ritualId);
+    function rituals(
+        uint32 ritualId
+    )
+        external
+        view
+        returns (
+            address initiator,
+            uint32 initTimestamp,
+            uint32 endTimestamp,
+            uint16 totalTranscripts,
+            uint16 totalAggregations,
+            //
+            address authority,
+            uint16 dkgSize,
+            uint16 threshold,
+            bool aggregationMismatch,
+            //
+            IEncryptionAuthorizer accessController,
+            BLS12381.G1Point memory publicKey,
+            bytes memory aggregatedTranscript,
+            IFeeModel feeModel
+        )
+    {
+        Ritual storage ritual = storageRitual(ritualId);
+        initiator = ritual.initiator;
+        initTimestamp = ritual.initTimestamp;
+        endTimestamp = ritual.endTimestamp;
+        totalTranscripts = ritual.totalTranscripts;
+        totalAggregations = ritual.totalAggregations;
+        authority = ritual.authority;
+        dkgSize = ritual.dkgSize;
+        threshold = ritual.threshold;
+        aggregationMismatch = ritual.aggregationMismatch;
+        accessController = ritual.accessController;
+        publicKey = ritual.publicKey;
+        aggregatedTranscript = ritual.aggregatedTranscript;
+        feeModel = ritual.feeModel;
     }
 
     // for backward compatibility
@@ -152,7 +187,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     }
 
     function getInitiator(uint32 ritualId) external view returns (address) {
-        return rituals(ritualId).initiator;
+        return storageRitual(ritualId).initiator;
     }
 
     function getTimestamps(
@@ -377,7 +412,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     }
 
     function getAuthority(uint32 ritualId) external view returns (address) {
-        return rituals(ritualId).authority;
+        return storageRitual(ritualId).authority;
     }
 
     function postAggregation(
--- a/contracts/contracts/coordination/Coordinator.sol
+++ b/contracts/contracts/coordination/Coordinator.sol
@@ -139,7 +139,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     }
 
     function rituals(
-        uint32 ritualId
+        uint256 ritualId // uint256 for backward compatibility
     )
         external
         view
@@ -161,7 +161,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
             IFeeModel feeModel
         )
     {
-        Ritual storage ritual = storageRitual(ritualId);
+        Ritual storage ritual = storageRitual(uint32(ritualId));
         initiator = ritual.initiator;
         initTimestamp = ritual.initTimestamp;
         endTimestamp = ritual.endTimestamp;
--- a/deployment/artifacts/lynx.json
+++ b/deployment/artifacts/lynx.json
@@ -1,7064 +1,7064 @@
 {
-  "11155111": {
-    "LynxStakingToken": {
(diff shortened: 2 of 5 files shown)


deployment/params.py
```diff
@@ -615,12 +619,15 @@ def _deploy_proxy(
         )
         return contract_type_container.at(proxy_contract.address)
 
-    def upgrade(self, container: ContractContainer, proxy_address, data=b'') -> ContractInstance:
+    def upgrade(self, container: ContractContainer, proxy_address, data=b"") -> ContractInstance:
+        implementation = self.deploy(container)
+        # TODO: initialize taco app implementation too
+        return self.upgradeTo(implementation, proxy_address, data)
```
derekpierre, line 625:
```suggestion
        implementation = self.deploy(container)
        # upgrade proxy to implementation
        return self.upgradeProxyTo(implementation, proxy_address, data)
```

scripts/lynx/upgrade_coordinator.py
```diff
@@ -18,12 +19,22 @@ def main():
     deployer = Deployer.from_yaml(filepath=CONSTRUCTOR_PARAMS_FILEPATH, verify=VERIFY)
     instances = contracts_from_registry(filepath=ARTIFACTS_DIR / "lynx.json", chain_id=80002)
 
-    coordinator = deployer.upgrade(
-        project.Coordinator, instances[project.Coordinator.contract_type.name].address
+    implementation = deployer.deploy(project.Coordinator)
+    # encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
+    encoded_initializer_function = b""
```
derekpierre, line 24: Do we no longer need the commented line above?

vzotova: it's needed as one time thing, I didn't want to remove it to have it as example for other network including mainnet

derekpierre: Ah ok. Makes sense.

derekpierre, line 24:
```suggestion
    implementation = deployer.deploy(project.Coordinator)
    # `initializeNumberOfRituals` was used for the original upgrade, but should NOT be 
    #  used for subsequent upgrades of Coordinator
    # encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
    encoded_initializer_function = b""
```

derekpierre: Made a suggestion for a comment to add about that - you can edit/adjust as you see fit.

derekpierre (approved):
🎸

scripts/lynx/upgrade_coordinator.py
vzotova: ✔️

manumonti (approved):
LGTM! 👌

scripts/lynx/upgrade_coordinator.py
```diff
@@ -18,12 +19,24 @@ def main():
...
     instances = contracts_from_registry(filepath=ARTIFACTS_DIR / "lynx.json", chain_id=80002)
 
-    coordinator = deployer.upgrade(
-        project.Coordinator, instances[project.Coordinator.contract_type.name].address
+    implementation = deployer.deploy(project.Coordinator)
+    # `initializeNumberOfRituals` was used for the original upgrade, but should NOT be
+    #  used for subsequent upgrades of Coordinator
+    # encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
+    encoded_initializer_function = b""
+    coordinator = deployer.upgradeTo(
```
cygnusv, line 27: Can you just still use `upgrade()`, not deploying the implementation explicitly above?

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -138,8 +138,43 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+            //
+            address authority,
+            uint16 dkgSize,
+            uint16 threshold,
+            bool aggregationMismatch,
+            //
+            IEncryptionAuthorizer accessController,
+            BLS12381.G1Point memory publicKey,
+            bytes memory aggregatedTranscript,
+            IFeeModel feeModel
```
cygnusv, line 161: What about we create a dummy struct for this? Something like `RitualView` or along those lines.

deployment/artifacts/lynx.json
```diff
@@ -1,7057 +1,7064 @@
 {
-    "11155111": {
```
cygnusv, line 2: There seems to be an issue here with the number of spaces for tabs, which is producing a huge diff. I'd prefer if the same indentation convention is preserved, so change history is better understood.

cygnusv (changes requested):
My only strong comment is regarding the JSON registry format, but apart from it looks good to me! The rest it's just nitpicking.

deployment/artifacts/lynx.json
derekpierre: I think this was `pre-commit` job kicking in. 

Initially, our registries weren't subject to this linter update from pre-commit, and then once the pre-commit linter job kicked in, we didn't want to have a massive diff, and so we continued to ignore the linter update.

Some options:
- Modify the pre-commit linter ignore these files
- Update the file at some point (now or later) based on the linter, and continue to let the linter ensure the format remains the same. If we update to the linter's format, we should probably also modify this line - https://github.com/nucypher/nucypher-contracts/blob/main/deployment/registry.py#L20 - to use indent 2.
...

contracts/contracts/coordination/Coordinator.sol
vzotova: I doubt it will look any better, but if you want I can do that

scripts/lynx/upgrade_coordinator.py
vzotova: ✔️

deployment/params.py
```diff
@@ -633,6 +633,13 @@ def _deploy_proxy(
         return contract_type_container.at(proxy_contract.address)
 
     def upgrade(self, container: ContractContainer, proxy_address, data=b"") -> ContractInstance:
+        implementation = self.deploy(container)
+        # upgrade proxy to implementation
+        return self.upgradeProxyTo(implementation, proxy_address, data)
```
derekpierre, line 638: Is this supposed to be:

```suggestion
        return self.upgradeTo(implementation, proxy_address, data)
```

derekpierre: Not that I'm against the new function name if you want to change the name of the actual function.

deployment/artifacts/lynx.json
derekpierre: Side note, if we go with keeping the original format, @vzotova you can use the `normalize_registry` script to revert the file back to what it was before the linter kicked in.

vzotova: ✔️

deployment/params.py
vzotova: ✔️

contracts/contracts/coordination/Coordinator.sol
cygnusv: If you don't think it's worth it, then it's fine, let's keep it this way

deployment/artifacts/lynx.json
derekpierre: Filed https://github.com/nucypher/nucypher-contracts/issues/319

cygnusv (approved):

PR #326, opened 2024-09-04
Upgrade tapir Coordinator

vzotova:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/scripts/tapir/upgrade_coordinator.py
+++ b/scripts/tapir/upgrade_coordinator.py
@@ -19,8 +19,13 @@ def main():
     deployer = Deployer.from_yaml(filepath=CONSTRUCTOR_PARAMS_FILEPATH, verify=VERIFY)
     instances = contracts_from_registry(filepath=ARTIFACTS_DIR / "tapir.json", chain_id=80002)
 
-    coordinator = deployer.upgrade(
-        project.Coordinator, instances[project.Coordinator.contract_type.name].address
+    implementation = deployer.deploy(project.Coordinator)
+    encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
+    encoded_initializer_function = b""
+    coordinator = deployer.upgradeTo(
+        implementation,
+        instances[project.Coordinator.contract_type.name].address,
+        encoded_initializer_function,
     )
 
     deployments = [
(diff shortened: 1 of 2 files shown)


scripts/tapir/upgrade_coordinator.py
```diff
@@ -19,8 +19,13 @@ def main():
     deployer = Deployer.from_yaml(filepath=CONSTRUCTOR_PARAMS_FILEPATH, verify=VERIFY)
     instances = contracts_from_registry(filepath=ARTIFACTS_DIR / "tapir.json", chain_id=80002)
 
-    coordinator = deployer.upgrade(
-        project.Coordinator, instances[project.Coordinator.contract_type.name].address
+    implementation = deployer.deploy(project.Coordinator)
+    encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
```
derekpierre, line 23: I know functionally it is the same, but should we comment this out now that the first update is completed, like we did for `lynx`?

vzotova: hmm, good question, I don't know, I really used those lines, and each upgrade is individual

derekpierre: Just to make sure I understand correctly. For the following lines:

```python
    encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
    encoded_initializer_function = b""
```

The 1st is originally what you ran the upgrade script with, and now moving forward for any subsequent upgrade only the 2nd should be used...

vzotova: you are correct with one addition: 2nd would be only in case there are no other reinitializer functions that need to be called

manumonti (approved):

scripts/tapir/upgrade_coordinator.py
derekpierre: Right.

What about something like the following:
```suggestion
    # latest reinitializer function used for most recent upgrade - reinitializer(2)
    # encoded_initializer_function = implementation.initializeNumberOfRituals.encode_input()
```

vzotova: ✔️

Issue #328, opened 2024-09-10
Static gas cost for ReimbursementPool doesn't include calldata cost


cygnusv:
Currently, `ReimbursementPool` calculates gas cost to reimburse as the sum of the gas spent (as measured by the caller) and the `staticGas`:
https://github.com/nucypher/nucypher-contracts/blob/48e71c81366858f63a09c2b566572ac63d5ab883/contracts/contracts/coordination/ReimbursementPool.sol#L64

`staticGas`, in turn, is defined as follows:
https://github.com/nucypher/nucypher-contracts/blob/48e71c81366858f63a09c2b566572ac63d5ab883/contracts/contracts/coordination/ReimbursementPool.sol#L19-L22

The base transaction cost is 21,000 gas, while the refund function gas can be computed at compile time. However, this is not taking into consideration that the calldata cost is actually dynamic and depends on the size in bytes; this cost is defined by https://eips.ethereum.org/EIPS/eip-2028 and https://eips.ethereum.org/EIPS/eip-2930.md as:

```
calldata cost = 21000 + countNonZeroBytesInCalldata * 16 + countZeroBytesInCalldata * 4 + countAccountAccessList * 2400 + countAccountStorageAccessList * 1900
```

For small calldata, we can approximate it to 21000, but for posting transcripts or aggregates (which may be in the order of KBs) it's an important cost. For example, a normal `postTranscript` call has a calldata cost of approximately 81000 gas (including the base 21000).

For our particular case, my recommendation would be to add an approximation of the formula above to the `processReimbursement()` function, since in that place we have access to `msg.data` (i.e., the calldata) and we can simply query `msg.data.length`. This tell us nothing about the number of (non-)zero bytes, but for simplicity we can assume the number of zero bytes is fixed (e.g., due to padding there won't be more than, say, 128 zero bytes). With respect to access list, we can assume it's not used (since we don't use them). So the formula I propose is:

```
calldata_size = msg.data.length
calldata cost = 21000 + (calldata_size - 128) * 16 + 128 * 4
```

A final note: calldata cost calculations may change in the future (see e.g. https://eips.ethereum.org/EIPS/eip-7623).

PR #334, opened 2024-09-16
Make Coordinator aware of transcript size

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
- Transcript size can now be deterministically computed from DKG size and threshold. Coordinator validates this accordingly. This is a tiny step in the direction of on-chain validation.
- Add calldata dynamic costs as part of the gas reimbursement calculation. Note that since we validate transcript sizes, this mechanism can't be abused by posting huge transcripts.

**Issues fixed/closed:**
* Closes https://github.com/nucypher/nucypher-contracts/issues/328

**Why it's needed:**

**Notes for reviewers:**


vzotova (approved):

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -385,6 +385,13 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         return keccak256(abi.encode(nodes));
     }
 
+    function expectedTranscriptSize(
+        uint16 dkgSize,
+        uint16 threshold
+    ) public pure returns (uint256) {
+        return 40 + (dkgSize + 1) * BLS12381.G2_POINT_SIZE + threshold * BLS12381.G1_POINT_SIZE;
```
KPrasch, line 392: Just a nitpick but `40` is the only unnamed component of this calculation.  Perhaps it can be assigned to a constant similar to the BLS point sizes?

KPrasch (approved):

tests/test_coordinator.py
```diff
@@ -8,41 +7,13 @@
...
 from web3 import Web3
 
+from tests.conftest import ONE_DAY, gen_public_key, generate_transcript, RitualState
+
 TIMEOUT = 1000
 MAX_DKG_SIZE = 31
 FEE_RATE = 42
 ERC20_SUPPLY = 10**24
 DURATION = 48 * 60 * 60
-ONE_DAY = 24 * 60 * 60
```
derekpierre, line 16: Thanks for this cleanup!

derekpierre (approved):
🎸 .

Did you also want to verify your updated Coordinator contract on lynx, with our updated verify script? 😉

contracts/contracts/coordination/Coordinator.sol
cygnusv: You're right, but the thing is that I don't understand well enough where that 40 comes from. I mean, it's a fixed overhead that comes from rust serialization but I don't have a full explanation for it. I'll update this when I have more info.

Issue #284, opened 2024-07-04
Tidy up testing fixtures


theref:
A lot of fixtures are being duplicated and can be tidied into `conftest.py`

PR #339, opened 2025-01-15
StandardSubscription

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
* Generalizes `BqETHSubscription` to `StandardSubscription`
* New subscription parameters (see https://docs.taco.build/fees/mainnet-fees)
* `GlobalAllowList` is now upgradeable
* Deployed new `GlobalAllowList` and new `StandardSusbcription` to Polygon Mainnet (including proxies and `ProxyAdmins`)
* Some updates to mainnet registry (there's some caveats that @derekpierre  raised but that we don't need to decide now)

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


derekpierre:
Since the name of the subscription contract has changed:
- we should move the `BqETHSubscription` entry from the `mainnet` contract registry - https://github.com/nucypher/nucypher-contracts/blob/main/deployment/artifacts/mainnet.json#L1799 - to a `bqeth.json` file and rename the entry to `StandardSubscription`.
- we should rename the entry in the `tapir` contract registry - https://github.com/nucypher/nucypher-contracts/blob/main/deployment/artifacts/tapir.json#L2338, and the `lynx` contract registry - https://github.com/nucypher/nucypher-contracts/blob/main/deployment/artifacts/lynx.json#L2553 - to `StandardSubscription`. I don't believe it was actually used by BqETH, it was just the name of the contract at the time.

deployment/artifacts/mainnet.json
```diff
@@ -1796,813 +1796,6 @@
         }
     },
     "137": {
-        "BqETHSubscription": {
```
derekpierre, line 1799: Do we need to save this entry in a separate bqeth.json file for deployment tracking purposes (with the rename of course), or is that tracked already somewhere else? Basically what you did for the marlin deployment in `new-subscription-mainnet.json`.

deployment/artifacts/new-susbcription-mainnet.json
```diff
@@ -0,0 +1,1039 @@
+{
```
derekpierre, line 1: Should we rename to `marlin.json` or something analogous?

deployment/constructor_params/mainnet/new-subscription.yml
```diff
@@ -0,0 +1,65 @@
+deployment:
```
derekpierre, line 1: Should we call this file `marlin-subscruption.yml`?

tests/test_global_allow_list.py
```diff
@@ -91,6 +91,20 @@ def global_allow_list(project, deployer, coordinator):
     return contract
 
 
+@pytest.fixture()
+def upgradeable_global_allow_list(project, deployer, coordinator, oz_dependency):
```
derekpierre, line 95: Do we envision using an upgradeable global allow list moving forward by default? If so, should we just update the original `global_allow_list` fixture to use a proxied instance?

derekpierre (commented):
My comments are more related to code/artifact organization, but the contracts/scripts updates look good.

tests/test_global_allow_list.py
vzotova: I'd keep original global allow list intact and transform this upgradeable to the managed allow list after it will be ready.
wdyt?

deployment/constructor_params/mainnet/new-subscription.yml
cygnusv: Not necessarily, there's nothing in here that's specific to them. It's just the updated fees and parameters for every potential new subscription. In this case, we include as well logic to deploy a new GlobalAllowList (coincidentally, this happened with BqETH too) but that can be removed when we deploy another new subscription.

tests/test_global_allow_list.py
cygnusv: >  If so, should we just update the original global_allow_list fixture to use a proxied instance?

In tests I keep both versions and parametrize the test.

deployment/constructor_params/mainnet/new-subscription.yml
derekpierre: I guess we really did two separate actions in one script:
- deploy a new global allow list contract (only needed this time, but not unless there are future changes to global allow list).
- deploy a new subscription contract based on parameters.

Moving forward, a "new subscription" would only use the 2nd part with some potential tweaks to the values eg. duration. Just trying to think about our code maintenance moving forward - this params file and the associated script would need to remove the global allow list component the next time we do a subscription.

Not something to be done for this PR but some follow-up reorg can be done afterward eg. 
- separate params/script for updating global allow list implementation
- update "new subscription files" to not include anything with global allow list so that is more of a base that can be used for other adopters.

deployment/artifacts/new-susbcription-mainnet.json
derekpierre: Also, do we need `GlobalAllowList` in this file since it is already tracked in `mainnet.json`? I guess it is a remnant from the fact that the deployment included a fresh deployment of GlobalAllowList - but for the purposes of the subscription registry tracking for marlin, isn't really needed.

tests/test_global_allow_list.py
derekpierre: > In tests I keep both versions and parametrize the test.

Yep, just thinking that we don't really need both since the testing of either is the same, but not an issue if you want to keep both.

From the test's perspective, whether proxied or not doesn't functionally matter - but a proxied contract is what is deployed so we would match what is deployed. 

We duplicate this fixture in other places too which we should proxy:
- https://github.com/nucypher/nucypher-contracts/blob/main/tests/test_coordinator.py#L89
- https://github.com/nucypher/nucypher-contracts/blob/2085cab2e1a053372b002c80160a75185f15f9ae/tests/test_standard_subscription.py#L95

There is an issue for cleaning up duplicated fixtures across `nucypher-contracts`, and this can be addressed then - https://github.com/nucypher/nucypher-contracts/issues/284

cygnusv:
Let's address artifacts reorganization in a subsequent PR. I don't have enough clarity of what's the best way moving forward, particularly since I think there's reliance from code in nucypher (see https://github.com/nucypher/nucypher/issues/3503) and nucypher-contracts (e.g. https://github.com/nucypher/nucypher-contracts/blob/main/scripts/manage_subscription.py) on the presence of specific entries in the registry. To be clear, I'm not advocating to keep these in the registry (I also agree that the registry should only be used to keep protocol specific contracts), but these dependencies should be removed first.

tests/test_global_allow_list.py
cygnusv: Agreed, let's address that as part of #284

derekpierre (approved):
🎸

vzotova (approved):

PR #341, opened 2025-01-20
Allow list with cohort and auth admin roles

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


tests/test_managed_allow_list.py
```diff
@@ -1,160 +1,101 @@
...
+from eth_account.messages import encode_defunct
+from web3 import Web3
 
 
 @pytest.fixture(scope="module")
-def deployer(accounts):
-    return accounts[0]
+def initiator(accounts):
+    initiator_index = 1
+    assert len(accounts) >= initiator_index
```
derekpierre, line 12: I know you are being complete here 😄 , but our ape-config generates 40 test accounts. I'm ok with this function not checking the index, but that's up to you.

In any case, as part of https://github.com/nucypher/nucypher-contracts/issues/284 we will likely make these fixtures common, so fine to leave this for now.

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,97 @@
...
-        for (uint256 i = 0; i < addresses.length; i++) {
-            require(
-                authActions[ritualId] <
-                    subscription.authorizationActionsCap(ritualId, addresses[i]),
-                "Authorization cap exceeded"
-            );
-        }
+    function grantAuthAdminRole(uint32 ritualId, address account) external {
+        initializeCohortAdminRole(ritualId);
+        grantRole(ritualRole(ritualId, AUTH_ADMIN_BASE), account);
```
derekpierre, line 49: So just to confirm, if the ritual authority calls `grantRole()` directly without having first called `initializeCohortAdminRole()`, the call will fail.

But, if the ritual authority either:
- makes the two separate calls in here themselves directly
OR 
- calls `grantAuthAdminRole()`

then granting the role will work correctly.

derekpierre (commented):
So based on @cygnusv 's comment in Discord:
```
Maybe we can just add a flag to disallow setting auth admins, so with the flag off the contract simply behaves as a normal global allows list. The point would be to have a single type of contract
```

The "flag" at the moment, is really just either using the `GlobalAllowList` or the `ManagedAllowList` as the `accessController` for a Ritual. Correct?

tests/test_managed_allow_list.py
vzotova: more like I forgot to remove it 
✔️

derekpierre (approved):
Approving (assuming my clarifications were correct) - nice work 🎸 !

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,97 @@
...
-    /**
-     * @notice Checks if the sender has allowance to set authorizations
-     * @dev This function overrides the canSetAuthorizations modifier in the GlobalAllowList contract
-     * @param ritualId The ID of the ritual
-     */
-    modifier canSetAuthorizations(uint32 ritualId) override {
-        require(getAllowance(ritualId, msg.sender) > 0, "Only administrator is permitted");
-        _;
+    function ritualRole(uint32 ritualId, bytes32 role) public view returns (bytes32) {
+        return keccak256(abi.encodePacked(ritualId, role));
```
cygnusv, line 29: Awesome idea 💯 Let's go a step further and have two functions based on this:
```js
function cohortAdminRole(uint32 ritualId) public view returns (bytes32);
function authAdminRole(uint32 ritualId) public view returns (bytes32);
```
Also, can this function be `pure`? At first sight, I don't see why not 🤔

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,97 @@
...
-        }
+    function grantAuthAdminRole(uint32 ritualId, address account) external {
+        initializeCohortAdminRole(ritualId);
+        grantRole(ritualRole(ritualId, AUTH_ADMIN_BASE), account);
     }
 
     /**
-     * @notice Sets the administrator caps for a ritual
-     * @dev Only active rituals can set administrator caps
+     * @notice Checks if the sender is the authority of the ritual
```
cygnusv, line 53:
```suggestion
     * @notice Checks if the sender is an Authorization Admin of the ritual
```

cygnusv (commented):
Still digesting

contracts/contracts/coordination/ManagedAllowList.sol
vzotova: ✔️

contracts/contracts/coordination/ManagedAllowList.sol
vzotova: good idea!
✔️

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,102 @@
...
 
-import "../lib/LookupKey.sol";
+import "@openzeppelin-upgradeable/contracts/access/AccessControlUpgradeable.sol";
 import "./GlobalAllowList.sol";
-import "./Coordinator.sol";
-import {UpfrontSubscriptionWithEncryptorsCap} from "./subscription/Subscription.sol";
 
 /**
  * @title ManagedAllowList
  * @notice Manages a list of addresses that are authorized to decrypt ciphertexts, with additional management features.
```
cygnusv, line 10:
```suggestion
 * @notice Manages a list of addresses that are authorized to encrypt ciphertexts, with additional management features.
```

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,102 @@
...
-        bool value
-    ) internal override {
-        super._beforeSetAuthorization(ritualId, addresses, value);
-        for (uint256 i = 0; i < addresses.length; i++) {
-            require(
-                authActions[ritualId] <
-                    subscription.authorizationActionsCap(ritualId, addresses[i]),
-                "Authorization cap exceeded"
-            );
+    function initializeCohortAdminRole(uint32 ritualId) public {
```
cygnusv, line 44: When it is expected this function to be called? The way I see it, we could do it in the coordinator when the ritual is initiated. This shouldn't matter in the case of ritual failing, but if ritual succeeds, then this is already done from day 1.

tests/test_managed_allow_list.py
```diff
@@ -1,160 +1,95 @@
...
+    assert managed_allow_list.hasRole(auth_admin_role, deployer)
+    managed_allow_list.authorize(ritual_id, [deployer.address], sender=deployer)
 
-def test_add_administrators(managed_allow_list, authority, admin):
-    # Only authority can add administrators
-    with ape.reverts("Only cohort authority is permitted"):
-        managed_allow_list.addAdministrators(RITUAL_ID, [admin.address], ADMIN_CAP, sender=admin)
+    managed_allow_list.grantRole(auth_admin_role, initiator, sender=initiator)
+    with ape.reverts("Encryptor must be authorized by the sender first"):
+        managed_allow_list.deauthorize(ritual_id, [deployer.address], sender=initiator)
```
cygnusv, line 84: Here's some food for thought for a new test:
* Cohort admin authorizes auth admin A
* Cohort admin authorizes auth admin B
* A authorizes encryptor E
* B authorizes encryptor E
* A deauthorizes encryptor E

What should happen with E? Is E still authorized or not?

cc @arjunhassard @derekpierre @KPrasch

cygnusv (commented):
More comments. There's still a design issue we need to address before moving forward.

tests/test_managed_allow_list.py
derekpierre: Interesting scenario. Given that `E` is used by both `A` and `B`, what scenario could yield that? 🤔 . An encryption service of some sort? But it would be using the same signer for both `A` and `B`? Maybe the user encrypting for themselves on two different apps (`A` and `B`) using the same ritual...?

At first blush, it seems that `E` should still be authorized for `B`. It's unclear the relationship between `A` and `B`, and feels like without that knowledge the safest philosophy is that `A`'s actions should not affect B's.

Did you have existing thoughts here?

vzotova: 'B authorizes encryptor E' - that will fail with the current code, encryptor can be authorized only by one auth admin

contracts/contracts/coordination/ManagedAllowList.sol
vzotova: I'd avoid any changes in Coordinator to call it 
This function included in grantAuthAdminRole function, so it will be called anyway during first granting

contracts/contracts/coordination/subscription/Subscription.sol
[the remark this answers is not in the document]
vzotova: Yep, Old `Subscription` contract was there just as reference (and mostly to compile old ManagedAllowList)

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,102 @@
...
-     * @param _subscription The address of the subscription contract
      */
-    constructor(
-        Coordinator _coordinator,
-        UpfrontSubscriptionWithEncryptorsCap _subscription // TODO replace with IFeeModel subscription
-    ) GlobalAllowList(_coordinator) {
-        require(address(_subscription) != address(0), "Subscription cannot be the zero address");
-        subscription = _subscription;
+    constructor(Coordinator _coordinator) GlobalAllowList(_coordinator) {
+        _disableInitializers();
```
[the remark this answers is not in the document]
vzotova: just a habit, I'll remove it

contracts/contracts/coordination/ManagedAllowList.sol
vzotova: ✔️

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,102 @@
...
+        address authority = coordinator.getAuthority(ritualId);
+        require(authority != address(0), "Ritual is not initiated");
+        if (hasRole(cohortAdminRole, authority)) {
+            return;
         }
+        _setRoleAdmin(authAdminRole(ritualId), cohortAdminRole);
+        _grantRole(cohortAdminRole, authority);
+    }
+
+    function grantAuthAdminRole(uint32 ritualId, address account) external {
```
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/coordination/ManagedAllowList.sol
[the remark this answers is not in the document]
vzotova: ✔️

tests/test_managed_allow_list.py
cygnusv: Right, if B can't authorize E becasue it's already authorized by A, then the problem disappears. We need to think carefully if this is acceptable, but if so, then I don't have further comments.

derekpierre: Would the following scenario be a possibility? Two apps, `A` and `B`, each being authAdmins for the same ritual (use as part of using the same broader service) and wanting to control their own set of encryptors. The encryptor could be the user and the user utilizes the same wallet across the two apps?

vzotova: could you elaborate a bit more
so we have two auth admins? and same ritual? each of them can have own set of encryptors within paid subscription

derekpierre: Yes. Effectively, two auth admins for the same ritual. e.g. Ceramic pays for a ritual, `r`, for their developers, and then assigns two separate developers/application owners, `A` and `B`, as auth admins as part of their respective applications.

Both applications allow their individual users to be encryptors.  User, `U`, uses both apps and wants to use their EOA wallet as an encryptor...?

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,119 @@
...
         uint32 ritualId,
         address[] calldata addresses
-    ) external onlyCohortAuthority(ritualId) {
-        setAdministratorCaps(ritualId, addresses, 0);
+    ) external override canSetAuthorizations(ritualId) {
+        for (uint256 i = 0; i < addresses.length; i++) {
+            bytes32 lookupKey = LookupKey.lookupKey(ritualId, addresses[i]);
+            require(
+                authAdmins[msg.sender][lookupKey],
+                "Encryptor must be authorized by the sender first"
```
manumonti, line 115:
```suggestion
                "Encryptor has not been previously authorized by the sender"
```

WDYT? IMHO the original message seems to express that there previous step to do before deauthorize, instead of expressing the reason of why the deauthorization can not be done.

manumonti (changes requested):
Looks good to me. I only have some wording suggestions.

contracts/contracts/coordination/ManagedAllowList.sol
vzotova: a bit long but everything longer than 32 bytes are not good anyway so I'm ok with suggestion

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,119 @@
...
-        );
-        _;
+    function cohortAdminRole(uint32 ritualId) public pure returns (bytes32) {
+        return ritualRole(ritualId, COHORT_ADMIN_BASE);
     }
 
     /**
-     * @notice Checks if the sender has allowance to set authorizations
-     * @dev This function overrides the canSetAuthorizations modifier in the GlobalAllowList contract
+     * Prepares auth admin role identificator for the specific ritual
```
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,119 @@
...
-    ) GlobalAllowList(_coordinator) {
-        require(address(_subscription) != address(0), "Subscription cannot be the zero address");
-        subscription = _subscription;
+    function ritualRole(uint32 ritualId, bytes32 role) public pure returns (bytes32) {
+        return keccak256(abi.encodePacked(ritualId, role));
     }
 
     /**
-     * @notice Checks if the sender is the authority of the ritual
+     * Prepares cohort admin role identificator for the specific ritual
```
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,119 @@
...
+    constructor(Coordinator _coordinator) GlobalAllowList(_coordinator) {}
 
     /**
-     * @notice Sets the coordinator and subscription contracts
-     * @dev The coordinator and subscription contracts cannot be zero addresses
-     * @param _coordinator The address of the coordinator contract
-     * @param _subscription The address of the subscription contract
+     * Prepares role identificator for the specific ritual
+     * @param ritualId The ID of the ritual
+     * @param role Role identificator
```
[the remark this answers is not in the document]
vzotova: ✔️

contracts/contracts/coordination/ManagedAllowList.sol
```diff
@@ -2,149 +2,119 @@
...
-import "../lib/LookupKey.sol";
+import "@openzeppelin-upgradeable/contracts/access/AccessControlUpgradeable.sol";
 import "./GlobalAllowList.sol";
-import "./Coordinator.sol";
-import {UpfrontSubscriptionWithEncryptorsCap} from "./subscription/Subscription.sol";
 
 /**
  * @title ManagedAllowList
- * @notice Manages a list of addresses that are authorized to decrypt ciphertexts, with additional management features.
+ * @notice Manages a list of addresses that are authorized to encrypt ciphertexts, with additional management features.
```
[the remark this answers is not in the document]
vzotova: ✔️

manumonti (approved):

theref (approved):

theref (approved):

PR #342, opened 2025-01-21
IFeeModel/Subscription contract should be provided as a contract address

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
SubscriptionContracts/Fee Models are per adopter i.e. the address used is different every time. Therefore, parameters that take this value should take the value as a 

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/scripts/initiate_ritual.py
+++ b/scripts/initiate_ritual.py
@@ -1,11 +1,11 @@
 #!/usr/bin/python3
 
 import click
-from ape import chain
+from ape import Contract, chain
 from ape.cli import ConnectedProviderCommand, account_option, network_option
 
 from deployment import registry
-from deployment.constants import ACCESS_CONTROLLERS, FEE_MODELS, SUPPORTED_TACO_DOMAINS
+from deployment.constants import ACCESS_CONTROLLERS, SUPPORTED_TACO_DOMAINS
 from deployment.params import Transactor
 from deployment.types import ChecksumAddress, MinInt
 from deployment.utils import check_plugins, sample_nodes
@@ -38,8 +38,8 @@ from deployment.utils import check_plugins, sample_nodes
 @click.option(
     "--fee-model",
     "-f",
-    help="The name of a fee model contract.",
-    type=click.Choice(FEE_MODELS),
+    help="The address of the subscription fee model contract.",
+    type=ChecksumAddress(),
     required=True,
 )
 @click.option(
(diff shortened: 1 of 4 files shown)


vzotova (approved):

scripts/initiate_ritual.py
```diff
@@ -38,8 +38,8 @@
 @click.option(
     "--fee-model",
     "-f",
-    help="The name of a fee model contract.",
-    type=click.Choice(FEE_MODELS),
+    help="The address of the subscription fee model contract.",
```
cygnusv, line 41:
```suggestion
    help="The address of the fee model/subscription contract.",
```
At some point we should resolve terminology re: subscription vs feeModel. In this particular case, the option should emphasize it's a fee model.

scripts/initiate_ritual.py
```diff
@@ -111,10 +111,10 @@ def cli(
     access_controller_contract = registry.get_contract(
         domain=domain, contract_name=access_controller
     )
-    fee_model_contract = registry.get_contract(domain=domain, contract_name=fee_model)
+    fee_model_contract = Contract(fee_model)
 
-    # if using a subcription, duration needs to be calculated
-    if fee_model == "StandardSubscription":
+    # if using a subscription, duration needs to be calculated
+    if fee_model_contract.contract_type.name == "StandardSubscription":
```
cygnusv, line 117: Are you sure this will work? My understanding is that in L114, ape will instantiate `fee_model_contract` as a generic `Contract`, so I'm unsure that `fee_model_contract.contract_type.name` will be available for this check.

scripts/manage_subscription.py
```diff
@@ -56,9 +56,8 @@ def pay_subscription(account, network, domain, subscription_contract, encryptor_
     check_plugins()
     click.echo(f"Connected to {network.name} network.")
     transactor = Transactor(account=account)
-    subscription_contract = registry.get_contract(
-        contract_name=subscription_contract, domain=domain
-    )
+    subscription_contract = Contract(address=subscription_contract)
```
cygnusv, line 59: Similar to the comment above. There's no associated ABI, as opposed in the original code that had access to the registry. Perhaps this can work in certain local configurations, if ape has prior knowledge about deployed ABIs for a particular address, but we can't make this assumption.

scripts/manage_subscription.py
```diff
@@ -85,9 +84,7 @@ def pay_slots(account, network, domain, subscription_contract, encryptor_slots):
...
```
cygnusv, line 87: See comment above.

scripts/initiate_ritual.py
derekpierre: Yep. Tested it in `ape-console` and by running the command using Marlin's subscription contract and then quitting once the time came to sign a tx. You can try it out.

Using `Contract(...)` should automatically fetch the information from the `ape-etherscan` configured explorer, itf it can't get it locally - https://github.com/ApeWorX/ape/blob/v0.8.14/src/ape/managers/chain.py#L1339.

This has been further improved in newer Ape versions as well - https://docs.apeworx.io/ape/stable/userguides/contracts.html#from-any-address (not our current version, but shows that the pattern holds moving forward).

scripts/manage_subscription.py
derekpierre: See comment above.

cygnusv (approved):

PR #349, opened 2025-03-19
Upgrade for Subscription contract with extended period

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/coordination/subscription/BqETHSubscription.sol
+++ b/contracts/contracts/coordination/subscription/BqETHSubscription.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+pragma solidity ^0.8.0;
+
+import "./StandardSubscription.sol";
+
+/**
+ * @title BqETHSubscription
+ */
+contract BqETHSubscription is StandardSubscription {
+    /**
+     * @notice Sets the coordinator and fee token contracts
+     * @dev The coordinator and fee token contracts cannot be zero addresses
+     * @param _coordinator The address of the coordinator contract
+     * @param _accessController The address of the global allow list
+     * @param _feeToken The address of the fee token contract
+     * @param _adopterSetter Address that can set the adopter address
+     * @param _initialBaseFeeRate Fee rate per node per second
+     * @param _baseFeeRateIncrease Increase of base fee rate per each period (fraction of INCREASE_BASE)
+     * @param _encryptorFeeRate Fee rate per encryptor per second
+     * @param _maxNodes Maximum nodes in the package
+     * @param _subscriptionPeriodDuration Maximum duration of subscription period
+     * @param _yellowPeriodDuration Duration of yellow period
+     * @param _redPeriodDuration Duration of red period
+     */
+    constructor(
+        Coordinator _coordinator,
+        GlobalAllowList _accessController,
+        IERC20 _feeToken,
+        address _adopterSetter,
+        uint256 _initialBaseFeeRate,
+        uint256 _baseFeeRateIncrease,
+        uint256 _encryptorFeeRate,
+        uint256 _maxNodes,
+        uint32 _subscriptionPeriodDuration,
+        uint32 _yellowPeriodDuration,
+        uint32 _redPeriodDuration
+    )
+        StandardSubscription(
+            _coordinator,
+            _accessController,
+            _feeToken,
+            _adopterSetter,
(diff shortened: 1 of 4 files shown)


tests/test_bqeth_subscription.py
```diff
@@ -0,0 +1,213 @@
...
+(at your option) any later version.
+
+nucypher is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
```
manumonti, line 16: not relevant at all, but are we still putting this header?

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+pragma solidity ^0.8.0;
+
+import "./StandardSubscription.sol";
+
+/**
+ * @title BqETHSubscription
+ */
+contract BqETHSubscription is StandardSubscription {
```
manumonti, line 10: Out of curiosity: why do we need to implement a new contract for the adopter? Why not just deploy StandardSubscription with the constructor parameters set for this adopter?

manumonti (approved):

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: because period calculation based on immutable duration. So this update includes new duration and skipping 0 period. Otherwise it will consider 0 period with length of year instead of 6 months

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -0,0 +1,62 @@
...
+        )
+    {}
+
+    /// @dev use `upgradeAndCall` for upgrading together with re-initialization
+    function reinitialize() external reinitializer(2) {
+        startOfSubscription = uint32(block.timestamp);
+    }
+
+    function getCurrentPeriodNumber() public view override returns (uint256) {
+        return super.getCurrentPeriodNumber() + 1;
```
cygnusv, line 60: Just brainstorming here: what if we set startOfSubscription 6 months before the original date? That way
```
    startOfSubscription  -= 31_556_952 / 2;  // Half a year
```
and then we wouldn't have to tweak period numbers?

tests/test_bqeth_subscription.py
cygnusv: Let's continue doing it, we're still under the same GH organization.

contracts/contracts/coordination/subscription/BqETHSubscription.sol
vzotova: ✔️

contracts/contracts/coordination/subscription/BqETHSubscription.sol
```diff
@@ -0,0 +1,58 @@
...
+            _maxNodes,
+            _subscriptionPeriodDuration,
+            _yellowPeriodDuration,
+            _redPeriodDuration
+        )
+    {}
+
+    /// @dev use `upgradeAndCall` for upgrading together with re-initialization
+    function reinitialize() external reinitializer(2) {
+        startOfSubscription -= 31_556_952 / 2; // Half a year;
```
cygnusv, line 56:
```suggestion
        startOfSubscription -= 31_536_000 / 2; // Half a year;
```
Sorry, the other number was wrong. Never just copy and paste the first number Google gives you 🙈

vzotova: I saw it, and checked so it's 365.25 basically which was totally fine by me. But ok, I'll change it to 365

cygnusv:
We're closing this PR for the moment due to a change of adopter needs. We can always open it later. Thanks for the hard work @vzotova !

PR #358, opened 2025-04-08
Upgrade Coordinator on Tapir

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
When originally deploying, had some issues generating the contract registry. Updating the dependencies fixed the issue, but led to some test failures based on an upgrade of the `ape` version being used - those have now been fixed.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


scripts/tapir/upgrade_coordinator.py
```diff
@@ -19,14 +19,17 @@ def main():
     deployer = Deployer.from_yaml(filepath=CONSTRUCTOR_PARAMS_FILEPATH, verify=VERIFY)
     instances = contracts_from_registry(filepath=ARTIFACTS_DIR / "tapir.json", chain_id=80002)
 
-    implementation = deployer.deploy(project.Coordinator)
+    # implementation = deployer.deploy(project.Coordinator)
```
manumonti, line 22: Is this line being commented out intentional? Don't we need to deploy the implementation contract first?

tests/test_global_allow_list.py
```diff
@@ -121,7 +121,8 @@ def initiate_ritual(coordinator, fee_model, erc20, authority, nodes, allow_logic
 
 # Using normal and upgradeable versions of global_allow_list.
 # Since both are fixtures, we need a small workaround to use them as parameters.
-# See https://engineeringfordatascience.com/posts/pytest_fixtures_with_parameterize/#using-requestgetfixturevalue-
+# See https://engineeringfordatascience.com/posts/
+# pytest_fixtures_with_parameterize/#using-requestgetfixturevalue-
```
manumonti, line 125: Is this line be splitting into two deliberately? or this was the formatter?

```suggestion
# See https://engineeringfordatascience.com/posts/pytest_fixtures_with_parameterize/#using-requestgetfixturevalue-
```

manumonti (dismissed):
I left some questions

scripts/tapir/upgrade_coordinator.py
derekpierre: It's intentional - the implementation gets deployed as part of the `deployer.upgrade` call - https://github.com/derekpierre/nucypher-contracts/blob/tapir-new-abi/deployment/params.py#L635.

That implementation line was done before and can be used with `deployer.upgradeTo()` which was something we did in the past but is no longer applicable hence commented out. You'll see the same thing in the `lynx/upgrade_coordinator.py` file which David used for upgrading the lynx Coordinator.

manumonti (approved):

PR #360, opened 2025-04-09
Remove duplicated ABI entries used for registry

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
I think the problem was that both the ABI for the proxy contract and the underlying implementation contract were being used leading to duplicated ABIs, and additional entries from the proxy contract ABI.

Instead we should get the abi from the underlying implementation contract and not the proxy contract. That removes the proxy ABIs part of the problem.

There are still duplicates, so these are proactively skipped.

**Issues fixed/closed:**
> - Fixes #...

Related to https://github.com/ApeWorX/ape/issues/2569

**Why it's needed:**

**Notes for reviewers:**


cygnusv (approved):

derekpierre:
> Seems like there will be a fix for this in 0.8.31 - https://discord.com/channels/922917176040640612/938897808147247186/1359905965083721800
> 
> We still need the code for updating the implementation instance from the proxy, but once we switch to 0.8.31 (when released) we will no longer need the duplicate check code. Tested locally with the latest from `ape`::`main` and a lynx upgrade and it worked without the duplicate check code.

ape 0.8.31 didn't solve the problem. In any case, not spending more time on this for now. Our workaround continues to work correctly.

We can merge this for now, and move on.

vzotova (approved):

derekpierre:
> ape 0.8.31 didn't solve the problem. In any case, not spending more time on this for now. Our workaround continues to work correctly.

Ape folks indicated that there some artifacts needed to be cleared before the changes would have an effect. Once that was done, the problem went away!

PR #362, opened 2025-04-21
Fix ci

manumonti:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
`deploy_child.py` script was failing on CI.

This change will fix it by using `autosign` parameter instead of the deprecated `non_interactive`.

Non_interactive function was removed on 06f56c58a55b394281db6ee1902bf0f5c1e0d44d.


derekpierre (approved):
🎸

PR #366, opened 2025-04-24
Signing POC

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [x] 3

**What this does:**
Initial Signing infrastructure:
- `SigningCoordinator` on ethereum
   - Signing ritual ceremony for establishing signing cohorts
- `SigningCoordinatorChild` contracts on subsequent L2s
   - each child deploys and tracks multisig contracts for the relevant cohort id
- Contract deployment scripts 
- Conditions are per-chain for signing cohorts

_I'm looking to deploy these contracts to `lynx`. This is a good base checkpoint for merging; not all functionality is implemented but subsequent PRs can be created into the `signing` epic._

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


contracts/contracts/coordination/OpL1Sender.sol
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * @param _dispatcher The address of the dispatcher contract.
+     * @param _messenger The address of the CrossDomainMessenger contract.
+     * @param _gasLimit The gas limit for the message.
+     */
+    constructor(address _dispatcher, address _messenger, uint32 _gasLimit) {
+        dispatcher = _dispatcher;
+        messenger = _messenger;
+        gasLimit = _gasLimit;
```
vzotova, line 32: Gas limit is a variable so it should go to initialize method or become immutable

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -0,0 +1,377 @@
...
+        uint32 indexed cohortId,
+        address indexed provider,
+        bytes signature
+    );
+    event SigningCohortDeployed(uint32 indexed cohortId, uint256 chainId);
+    event SigningCohortConditionsSet(uint32 indexed cohortId, uint256 chainId, bytes conditions);
+
+    event TimeoutChanged(uint32 oldTimeout, uint32 newTimeout);
+    event MaxCohortSizeChanged(uint16 oldSize, uint16 newSize);
+    event DispatcherChanged(address oldDispatcher, address newDispatcher);
```
vzotova, line 61: would be slightly better if put events before all variables/immutables or after, just for better reading

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -0,0 +1,377 @@
...
+
+    TACoApplication public immutable application;
+    SigningCoordinatorDispatcher public signingCoordinatorDispatcher;
+
+    uint96 private immutable minAuthorization;
+
+    uint32 public timeout;
+    uint16 public maxCohortSize;
+
+    SigningCohortParticipant internal __sentinelSigner;
```
vzotova, line 79: just for unification - adding gap (similar as in Coordinator) will help in case any contract will inherit this one

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -0,0 +1,377 @@
...
+    // TODO: this should be removed post-testing
+    function setDispatcher(
+        SigningCoordinatorDispatcher dispatcher
+    ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(address(dispatcher).code.length > 0, "Dispatcher must be contract");
+        emit DispatcherChanged(address(signingCoordinatorDispatcher), address(dispatcher));
+        signingCoordinatorDispatcher = dispatcher;
+    }
+
+    function _isCohortActive(SigningCohort storage _cohort) internal view returns (bool) {
```
vzotova, line 118: no need in underscore because function has different parameters compering to other `isCohortActive`
same about `_cohort` - there is no other variable with the same name so no need to cover it with underscore

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -0,0 +1,377 @@
...
+             *   - All transcripts and all aggregations
+             *   - Still within the deadline
+             */
+            revert("Ambiguous ritual state");
+        }
+    }
+
+    function getSigningCoordinatorChild(uint256 chainId) external view returns (address) {
+        address child = signingCoordinatorDispatcher.getSigningCoordinatorChild(chainId);
+        require(child != address(0), "No child application found");
```
vzotova, line 374: considering that this is a view maybe no need to revert tx? basically outside caller will decide what to do with that if there is no child?

vzotova (commented):
EPIC!

contracts/contracts/coordination/OpL1Sender.sol
derekpierre: ✔️

contracts/contracts/coordination/SigningCoordinator.sol
derekpierre: ✔️

contracts/contracts/coordination/SigningCoordinator.sol
derekpierre: Good point - ✔️

contracts/contracts/coordination/SigningCoordinator.sol
derekpierre: I added the underscore to the function name because it is an internal function. Is that the convention or was I mistaken?

vzotova: yeah, it is underscore only for cases when there is function/variable with that name.

contracts/contracts/coordination/SigningCoordinator.sol
vzotova: there is no strict rules, it could be any number, we usually start from 20

contracts/contracts/coordination/SigningCoordinator.sol
derekpierre: ✔️

contracts/contracts/coordination/SigningCoordinator.sol
vzotova: looks good

manumonti (approved):
This is a great work 🔥🔥🔥

I know this is just the start of the work / PoC, but something to consider in subsequents PRs is to add a little description of the contracts at the beginning of each one so it could be easier to follow what is the purpose of each.

I could help with that.

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -0,0 +1,379 @@
...
+    struct SigningCohort {
+        address initiator;
+        uint32 initTimestamp;
+        uint32 endTimestamp;
+        // TODO: what's the point of authority if we have a single global cohort?
+        address authority;
+        uint16 totalSignatures;
+        uint16 numSigners;
+        uint16 threshold;
+        SigningCohortParticipant[] signers;
```
cygnusv, line 57: I remember we had early discussions for the Coordinator about making this an array or a mapping. Each of them have their pros and cons, but I think, in retrospective, I'd have made this a mapping. One of the pros is that this completely removes the need for the whole pre/post sentinel gaps and other concerns that originate from potential changes to the `SigningCohortParticipant` struct.

contracts/contracts/coordination/OpL1Sender.sol
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @notice Sends data to the L2 receiver contract via the messenger.
+     * @param target The address of the target contract on L2.
+     * @param data The data to be sent to the target contract.
+     */
+    function sendData(address target, bytes calldata data) external {
+        require(dispatcher == msg.sender, "Unauthorized caller");
+        bytes memory payload = abi.encode(target, data);
+        bytes memory callData = abi.encodeWithSelector(IL2Receiver.recvData.selector, payload);
+        ICrossDomainMessenger(messenger).sendMessage(l2Receiver, callData, gasLimit);
```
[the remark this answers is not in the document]
derekpierre: ✔️  - fixed

contracts/contracts/coordination/SigningCoordinator.sol
derekpierre: The challenge with using a mapping is that it can't be iterated over. So a function like the following would be difficult:
```solidity
    function getSigners(uint32 cohortId) external view returns (SigningCohortParticipant[] memory) {
         ...
    }
```
Unless, of course, you’ve got some clever solidity workarounds in mind 😉 ?

cygnusv: Yep, not a Solidity hack per se, but if we use mappings then there would be no need to implement a `findSigner()` method, at the expense of dynamically constructing a `SigningCohortParticipant[]` array in `getSigners()` by iterating over an array of addresses (which we would have to store separately). So it's really a trade-off that we can discuss, but by no means a required change now.

PR #370, opened 2025-04-30
New Mainnet Coordinator ABI with breaking changes for nodes running nucypher <7.5

manumonti:
**This must be merged after #356 is merged**

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
The new Coordinator contract with breaking changes for nodes running nucypher < 7.5 has been deployed:

`0x4986b26f18ECA98474D85Ea583F99452983DCA1D`
https://polygonscan.com/address/0x4986b26f18ECA98474D85Ea583F99452983DCA1D#readContract

This is the ABI of the new contract deployment.

### DIFF
--- a/deployment/artifacts/mainnet.json
+++ b/deployment/artifacts/mainnet.json
@@ -3131,17 +3131,17 @@
                     "stateMutability": "view",
                     "inputs": [
                         {
-                            "name": "ritualId",
+                            "name": "",
                             "type": "uint32",
                             "internalType": "uint32"
                         },
                         {
-                            "name": "evidence",
+                            "name": "",
                             "type": "bytes",
                             "internalType": "bytes"
                         },
                         {
-                            "name": "ciphertextHeader",
+                            "name": "",
                             "type": "bytes",
                             "internalType": "bytes"
                         }
@@ -3203,7 +3203,7 @@
                     "stateMutability": "view",
                     "inputs": [
                         {
-                            "name": "provider",
+                            "name": "",
                             "type": "address",
                             "internalType": "address"
                         }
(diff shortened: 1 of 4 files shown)


deployment/artifacts/mainnet.json
```diff
@@ -3666,9 +3666,9 @@
                     "outputs": []
                 }
             ],
-            "tx_hash": "0x7858c6f34297752433f772ef6b7fa7086608ac3207fc9df5d4d6136ed7ab9b1f",
-            "block_number": 70019855,
-            "deployer": "0x1591165F1BF8B73de7053A6BE6f239BC15076879"
+            "tx_hash": "0xbfdb1ba47830e495dcae80c93943c4a0be539f0e94f795f0330d8e33e6c025fb",
+            "block_number": 70938418,
+            "deployer": "0x0224B7B41E9204550b8B3Fdc1afd6200446576E8"
```
derekpierre, line 3671: This part shouldn't have changed for a proxy upgrade, no? Or am I mistaken (cc @KPrasch )

vzotova (approved):

deployment/artifacts/mainnet.json
cygnusv: IIRC, our current default is to override these values with the latest implementation data, but it's true that it's not technically necessary.

cygnusv (approved):
Looks good! A musing: I initially didn't intend to update the registry since one of the premises of this upgrade was that the change was going to break the logic but not the API, and as such, the ABI shouldn't change. There's some minor cosmetic ABI changes that purely stem from the linter complaining about named parameters that were unused, thus we ended up using unnamed parameters in some cases, but I would argue those are not real ABI changes. Or perhaps more correctly, these ABI changes don't have any impact on any ABI-consuming process, and hence, I didn't think in updating the registry.

However, now that I see this PR from @manumonti, it made me think that it's actually good practice to update registries with every upgrade, as part of a predefined process. Thanks for thinking about this @manumonti !

derekpierre (approved):
🎸

PR #378, opened 2025-05-16
Fix bug where proxy contract address was not correctly being determined

derekpierre:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
The logic for obtaining the proxy contract for a deployed contract was not correctly being determined. Consequently, for deployment params that did something like:
```
  - MockPolygonRoot:
      constructor:
        _rootApplication: $TACoApplication
 ```
and `$TACoApplication` was deployed as a proxied contract, the implementation address was used instead of the proxy.

There was an update, https://github.com/nucypher/nucypher-contracts/commit/a30418e06fdec37ef6e1b059807f0a123edc3f80, that was a part of https://github.com/nucypher/nucypher-contracts/pull/358, and changed as part of an update to the ape version being used, but it doesn't seem that the change was correct. While this code was incorrect, I don't believe it has been used because we've only been updating contracts since then, which doesn't utilize that code.

I've been testing the change with my signing infra work.

This was also a problem when used as part of proxy constructor parameters. For example:
```
    SigningCoordinatorChild:
      proxy:
        constructor:
          _data: $encode:initialize,$ThresholdSigningMultisigCloneFactory,$SigningCoordinatorDispatcher,$deployer
```

Where `SigningCoordinatorDispatcher` was deployed as a proxy - the implementation address was used, but this is now fixed.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/deployment/params.py
+++ b/deployment/params.py
@@ -419,7 +419,7 @@ class ProxyParameters:
                     contract_names=contract_names,
                     constants=constants,
                     contract_name=contract_name,
-                    check_for_proxy_instances=False,
+                    check_for_proxy_instances=True,
                 ),
             )
             contracts_proxy_info.update({contract_name: proxy_info})
--- a/deployment/params.py
+++ b/deployment/params.py
@@ -153,9 +153,11 @@ class ContractName(Variable):
 
         if self.check_for_proxy_instances:
             # check if contract is proxied - if so return proxy contract instead
-            proxy_info = chain.contracts.get_proxy_info(contract_instance.address)
-            if proxy_info:
-                return proxy_info.address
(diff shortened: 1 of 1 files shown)


manumonti (approved):
LGTM! thanks for the useful explanations in the PR description 🙌

KPrasch (commented):
Is this caused by a newer version of eth-ape?

vzotova (approved):

PR #390, opened 2025-06-04
Add return of error code (1) for initiate_ritual

manumonti:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
Up to now, the script initiate_dkg.sh, aimed to be used on GitHub workflows, doesn't return an error code even if the initiate_ritual ape script fails. So the Github actions seems to be executed properly when it is not.

With these changes, the Github workflow will appear as failed when the script fails.

### DIFF
--- a/.github/scripts/initiate_dkg.sh
+++ b/.github/scripts/initiate_dkg.sh
@@ -20,5 +20,10 @@ ape run initiate_ritual                           \
 --duration ${DURATION}                            \
 --excluded-nodes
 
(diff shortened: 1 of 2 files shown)


KPrasch (approved):

derekpierre (approved):
🎸

PR #392, opened 2025-06-10
Establish a rate_limit for etherscan calls

manumonti:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
The default rate_limit for free subscription to Etherscan if 5 calls per second.

This will avoid getting error responses from Etherscan:

`ERROR:    Response indicated failure: Max calls per sec rate limit
reached (5/sec)`

### DIFF
--- a/.github/workflows/heartbeat.yml
+++ b/.github/workflows/heartbeat.yml
@@ -33,10 +33,10 @@ jobs:
     - name: Initiate Ritual
       run: .github/scripts/initiate_dkg.sh
       env:
-
         # Secret environment variables (secrets)
         APE_ACCOUNTS_automation_PASSPHRASE: ${{ secrets.DKG_INITIATOR_PASSPHRASE }}
         WEB3_INFURA_API_KEY: ${{ secrets.WEB3_INFURA_API_KEY }}
+        WEB3_INFURA_API_SECRET: ${{ secrets.WEB3_INFURA_API_SECRET }}
         ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/heartbeat.yml
```diff
@@ -33,10 +33,10 @@ jobs:
     - name: Initiate Ritual
       run: .github/scripts/initiate_dkg.sh
       env:
-
         # Secret environment variables (secrets)
         APE_ACCOUNTS_automation_PASSPHRASE: ${{ secrets.DKG_INITIATOR_PASSPHRASE }}
         WEB3_INFURA_API_KEY: ${{ secrets.WEB3_INFURA_API_KEY }}
+        WEB3_INFURA_API_SECRET: ${{ secrets.WEB3_INFURA_API_SECRET }}
```
derekpierre, line 39: Do we need both of these?

manumonti: > Do we need both of these - API_SECRET and API_KEY ?

Yes. The API_KEY is the Infura API Key as we already know and the API_SECRET is a "password" that brings an additional security measure. By this way, when ape-etherscan plughin uses both, it is possible to override the allowlist on our `Github Actions` Infura API key:

image

derekpierre: This is the `Github Actions` API Key? Did you add a API_SECRET or was it there before? If so, will that cause actions from other repos to fail?

PR #397, opened 2025-06-16
Fix node removal from excluded nodes list

manumonti:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
When we add to the excluded nodes list a node that is not part of the round (maybe because the node is shut down, so the script is not considering it to be part of the heartbeat rond), the script returns an error since it is trying to remove something (`node`) from a list that is not containing that element.

This change fixes it.


derekpierre (approved):
🎸

PR #399, opened 2025-06-23
Set excluded nodes as GH var instead of secret

manumonti:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
There is no reason to have the list of excluded nodes of DKG heartbeat rituals as GH actions secrets instead of variables.

Also, it is easier to troubleshoot the workflow if this is not a secret because the logs will not hide this information.


derekpierre (approved):
🎸 . Looks good. I noticed that there is still a secret variable set for `EXCLUDED_NODES` for the repos - if that secret is no longer needed we should remove it so we don't keep stale secret variables.

cygnusv (approved):

PR #403, opened 2025-07-31
[WIP] Check if node is participating in ritual before leaving the network

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/coordination/TACoChildApplication.sol
+++ b/contracts/contracts/coordination/TACoChildApplication.sol
@@ -266,6 +266,22 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
     ) external view returns (uint96 allAuthorizedTokens, bytes32[] memory activeStakingProviders) {
         return getActiveStakingProviders(_startIndex, _maxStakingProviders, 0);
     }
+
+    function release(
+        address _stakingProvider
+    ) external override(ITACoRootToChild, ITACoChildToRoot) {
+        // check all rituals
+        for (uint32 i = 0; i < Coordinator(coordinator).numberOfRituals(); i++) {
+            if (
+                Coordinator(coordinator).isRitualActive(i) &&
+                Coordinator(coordinator).isParticipant(i, _stakingProvider)
+            ) {
+                // still part of the ritual
+                return;
+            }
+        }
+        rootApplication.release(_stakingProvider);
+    }
 }
 
 contract TestnetTACoChildApplication is AccessControlUpgradeable, TACoChildApplication {
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -169,6 +169,12 @@ contract TACoApplication is
     event Penalized(address indexed stakingProvider, uint256 penaltyPercent, uint64 endPenalty);
 
     /**
+     * @notice Signals that the staking provider was released
+     * @param stakingProvider Staking provider address
+     */
+    event Released(address indexed stakingProvider);
+
+    /**
      * @notice Signals that reward was reset after penalty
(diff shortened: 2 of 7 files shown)


contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -75,8 +84,16 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
         adjudicator = _adjudicator;
     }
 
+    function setActiveRituals(uint32[] memory _activeRituals) external {
```
derekpierre, line 87: Should there be a restriction on who can call this?

derekpierre (commented):
With the caveat that ritual formation is currently permissioned, I'm wondering about something:
- can a "released" staker ever be switched back to "unreleased" if it becomes a member of new cohort?
- should a "released" staker be excluded from `getActiveStakingProviders` used for sampling nodes for new cohorts?

Or are these concerns not applicable to supervised handover in the short-term, and applicable to non-supervised/permissionless handover in the longer term?

contracts/contracts/coordination/TACoChildApplication.sol
vzotova: thank you, missed it!

vzotova:
> With the caveat that ritual formation is currently permissioned, I'm wondering about something:
> 
> * can a "released" staker ever be switched back to "unreleased" if it becomes a member of new cohort?
> * should a "released" staker be excluded from `getActiveStakingProviders` used for sampling nodes for new cohorts?
> 
> Or are these concerns not applicable to supervised handover in the short-term, and applicable to non-supervised/permissionless handover in the longer term?

* I ignored "unrelease" on purpose because it's hard to make it work without proper staking contract. So yeah it's postponed. And node can't become a member of new cohort with "released" state (see `athorizedStake`)
* it's already excluded because `eligibleStake` will return 0 and node will be skipped in `getActive` method

contracts/contracts/coordination/TACoChildApplication.sol
vzotova: Contract is not ownable so I added `initializer` to restrict it

Muhammad-Altabba (approved):
The proposed changes looks good to me.
However, I have some confusion and doubts about the original data structure, particularly the root <=> child relation that is one-to-one (which sounds like it is meant to be 1-to-many) and things like the firing of the same event name with the same parameters in each of those 2 contracts...
I will try to learn more about this and what each class actually represents in our next meetings...

contracts/contracts/coordination/TACoChildApplication.sol
```diff
@@ -75,8 +84,16 @@ contract TACoChildApplication is ITACoRootToChild, ITACoChildApplication, Initia
         adjudicator = _adjudicator;
     }
 
+    function setActiveRituals(uint32[] memory _activeRituals) external reinitializer(2) {
```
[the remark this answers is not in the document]
derekpierre: Only part of initializer now.

contracts/contracts/coordination/TACoChildApplication.sol
derekpierre: ^ Do we need updates to the deployment scripts for the upgrade today?

vzotova: yes and no, we can execute two steps upgrade on testnet, for the mainnet - yeah we want slight change in upgrade script

PR #424, opened 2025-09-12
Implement penalization for outdated nodes

manumonti:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
So far, we don't check the nodes version when we are calculating staking rewards. These changes are aimed to check if the node is running a valid version in every DKG heartbeat.

`evaluate_heartbeat.py` script will be executed every Monday 3h later (ritual timeout) of DKG heartbeats and will evaluate if the rituals were executed properly and if the nodes are running a valid version.

**Issues fixed/closed:**
This will close https://github.com/nucypher/sprints/issues/237

**Notes for reviewers:**
Based over PR #425 

Example of the artifact that is the input for the script `heartbeat-rituals.json`:
```
{
    "1208": [
        "0xD680B01503CF95A489010dA7409980060991D096",
        "0xFCA99f5b5Ecf19343e309d8dbe79F15a0A389769"
    ],
    "1209": [
        "0xB98a82aB97171504bf9eA0aC62A3f34BffE5F510",
        "0xF9283aC13aEc8a2B60a56A82F1f5E4dA1742CCf8"
    ],
    "1210": [
        "0x0154C52ec5b6a3010758dDe78079589E67526767",
        "0x97d065B567cc4543D20dffaa7009f9aDe64d7E26"
    ],
    "1211": [
        "0xA829c2E33907Ab03d27d5dBb39eFFd8386112789",
        "0xc1268db05E7bD38BD85b2C3Fef80F8968a2c933A"
    ],
    "1212": [
        "0x378C89b5B93667A005856822e1b65e2d7C076373",
        "0xd517479ACAef258Baa3ba6A351e9C714a79d597e"
    ]
}
```

Example of the artifact `offenders.json` that the script generates:

```
{
    "0xFCA99f5b5Ecf19343e309d8dbe79F15a0A389769": {
        "ritual": "1208",
        "reasons": [
            "Node is unreachable",
            "Missing transcript"
        ],
        "version": "Node is unreachable",
        "operator": "0x7EBb87dFf0aFBB3c5851daE39963b065e79Cc230",
        "eth_balance": 28.384842033281775
    },
    "0xF9283aC13aEc8a2B60a56A82F1f5E4dA1742CCf8": {
        "ritual": "1209",
        "reasons": [
            "Missing transcript"
        ],
        "version": "7.6.0",
        "operator": "0x917D3c54eb00ee781423Fe63f46C54D0F5195E13",
        "eth_balance": 9.14500647800536
    },
    "0x0154C52ec5b6a3010758dDe78079589E67526767": {
        "ritual": "1210",
        "reasons": [
            "Missing transcript"
        ],
        "version": "7.6.0",
        "operator": "0x73644B912d38b7C78E375a6057EF7F3D31360D39",
        "eth_balance": 17.33583245813157
    },
    "0xd517479ACAef258Baa3ba6A351e9C714a79d597e": {
        "ritual": "1212",
        "reasons": [
            "Missing transcript"
        ],
        "version": "7.6.0",
        "operator": "0x91e4e651d796938A2fa56505ab150FE782dF2aec",
        "eth_balance": 32.50976530044608
    }
}
```

Example of the artifact `discord_message.txt` generated by the script in a separate comment below.


scripts/evaluate_heartbeat.py
```diff
@@ -67,13 +68,51 @@ def get_operator(staker_address: str, taco_application: ContractContainer) -> st
...
+                        f"https://{rest_url}/status/",
+                        params={"json": "true"},
+                        verify=False,
+                        timeout=30,
+                    )
+                    if node_status.status_code == 200:
+                        return network_data.get("version", UNREACHABLE)
+
+                except (requests.ConnectionError, requests.exceptions.ReadTimeout):
+                    return UNREACHABLE
```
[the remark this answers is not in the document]
derekpierre: Another option could be:

```suggestion
                    node_status.raise_for_status()
                    return node_status.json().get("version", UNREACHABLE)
                except requests.RequestException:
                    return UNREACHABLE
```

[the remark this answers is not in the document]
manumonti: Yeah. I like it.

Just  to confirm I'm understanding the issue correctly. This is intended to stop the for loop quickly, but the result eventually will be the same (`UNREACHABLE`) because the last line of the function is `return UNREACHABLE`. Am I right?

scripts/evaluate_heartbeat.py
```diff
@@ -67,13 +68,57 @@ def get_operator(staker_address: str, taco_application: ContractContainer) -> st
...
+def get_valid_versions() -> List[Version]:
+    """Fetches valid versions considering the update grace period."""
+    releases_url = "https://api.github.com/repos/nucypher/nucypher/releases"
+    releases_response = requests.get(releases_url).json()
+
+    valid_versions: List[Version] = []
+
+    # GitHub seems to return releases in published_at-descending-order
+    for i, release in enumerate(releases_response):
+        version = Version(release.get("tag_name"))
```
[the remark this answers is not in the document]
manumonti: Fixed in 5056130820966fc4f67c4fbf281c079de089b3bf

scripts/evaluate_heartbeat.py
```diff
@@ -67,13 +68,69 @@ def get_operator(staker_address: str, taco_application: ContractContainer) -> st
...
+            return []
+
+        if i == 0:
+            valid_versions.append(version)
+        else:
+            previous_release_date = datetime.fromisoformat(releases_response[i - 1]["published_at"])
+            deadline = previous_release_date + NODE_UPDATE_GRACE_PERIOD
+            # filter the versions that are considered valid to run
+            if deadline > datetime.now(tz=timezone.utc):
+                valid_versions.append(version)
```
derekpierre, line 131: I think you can also be more efficient here and break out of the loop once a non-valid version is encountered. If the release response is in descending order, then all other previous versions would also be invalid.

So something like:


```suggestion
            if deadline > datetime.now(tz=timezone.utc):
                valid_versions.append(version)
            else:
                # invalid, so all prior versions are also invalid
                break
```

scripts/evaluate_heartbeat.py
```diff
@@ -67,13 +68,69 @@ def get_operator(staker_address: str, taco_application: ContractContainer) -> st
...
+                f"⚠️  Failed to parse version from release {version_str}"
+                + " Not semantic versioning format?",
+                fg="red",
+            )
+            return []
+
+        if i == 0:
+            valid_versions.append(version)
+        else:
+            previous_release_date = datetime.fromisoformat(releases_response[i - 1]["published_at"])
```
derekpierre, line 127: Good catch!

derekpierre (commented):
Have you been able to test out the script?

manumonti:
> Have you been able to test out the script?

Kind of. Only with old DKG heartbeat rounds. I will be able to test it better next monday with a fresh heartbeat round

scripts/evaluate_heartbeat.py
```diff
@@ -294,65 +280,134 @@ def investigate_offenders(
...
+            offenders[address] = {"ritual": ritual_id, "reasons": [], "version": version}
+
+            # check if node is reachable and running a valid version
+            if version == UNREACHABLE:
+                offenders[address]["reasons"].append(UNREACHABLE)
+                click.secho(f"Node {address} is unreachable", fg="cyan")
+            else:
+                try:
+                    version = Version(version)
+                    if version not in valid_versions:
```
vzotova, line 356: would this include new version that are not part of any release yet? like experimental? or we don't really care about them?

vzotova (approved):

scripts/evaluate_heartbeat.py
derekpierre: This would only be for published releases - https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases

derekpierre: Oh I see what you mean. Like someone running an unreleased (but newer eg. v7.7.0) experimental build on mainnet when a heartbeat round occurs.

@manumonti maybe you could also do a check if the version is > latest release, and if so then that be ok...?

theref (approved):

scripts/evaluate_heartbeat.py
manumonti: I'm addressing this in a following PR

PR #426, opened 2025-09-19
Adds method to reorder participants

vzotova:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/tests/test_coordinator.py
+++ b/tests/test_coordinator.py
@@ -1248,22 +1248,48 @@ def test_reorder_participants(
     pairs = [[0, 3], [1, 2], [5, 10]]
 
     activate_ritual(nodes, coordinator, ritualID)
-    participant_1 = coordinator.getParticipantFromProvider(ritualID, nodes[0])
-    participant_2 = coordinator.getParticipantFromProvider(ritualID, nodes[1])
-    participant_3 = coordinator.getParticipantFromProvider(ritualID, nodes[2])
-    participant_4 = coordinator.getParticipantFromProvider(ritualID, nodes[3])
-    participant_6 = coordinator.getParticipantFromProvider(ritualID, nodes[5])
-    participant_11 = coordinator.getParticipantFromProvider(ritualID, nodes[10])
+
+    # Get old participants with getParticipants() for later comparison
+    old_participants = coordinator.getParticipants(ritualID)
+
+    # Get individual participants via getParticipantFromProvider()
+    participant_0 = coordinator.getParticipantFromProvider(ritualID, nodes[0])
+    participant_1 = coordinator.getParticipantFromProvider(ritualID, nodes[1])
+    participant_2 = coordinator.getParticipantFromProvider(ritualID, nodes[2])
+    participant_3 = coordinator.getParticipantFromProvider(ritualID, nodes[3])
+    participant_5 = coordinator.getParticipantFromProvider(ritualID, nodes[5])
+    participant_10 = coordinator.getParticipantFromProvider(ritualID, nodes[10])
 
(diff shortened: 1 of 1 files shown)


contracts/contracts/coordination/Coordinator.sol
```diff
@@ -69,6 +69,7 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
         address indexed departingParticipant,
         address indexed incomingParticipant
     );
+    event ParticipantsReordered(uint32 indexed ritualId, uint256 index1, uint256 index2);
```
cygnusv, line 72: Not sure if we need this event.

cygnusv (approved):
Looks good! Just a minor comment.
Question: should we merge this? Seems like a one-off thing, and we can merge it to an upstream branch without affecting the main commit history.

wdyt? @vzotova @derekpierre @KPrasch

manumonti (approved):
LGTM! 👌

derekpierre:
> Looks good! Just a minor comment. Question: should we merge this? Seems like a one-off thing, and we can merge it to an upstream branch without affecting the main commit history.
> 
> wdyt? @vzotova @derekpierre @KPrasch

Agreed. No need to merge since it is a one off thing.

I assume the next step after deploying this change and doing the reordering is basically to put Coordinator back to the way it was before (this change). So the overall change (reordering and putting Coordinator back to the original version) is a no-op in terms of code changes.

I suppose in the future if we ever needed to remember what we did for the reordering we can always look back at this PR, which I assume we'll simply close instead of merging? OR would we rather have it stored in some upstream branch/separate tag?

derekpierre (approved):
🎸

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -876,4 +877,30 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
     //     require(tokenBalance > 0, "Insufficient balance");
     //     token.safeTransfer(msg.sender, tokenBalance);
     // }
+
+    function reorderParticipants(
+        uint32 ritualId,
+        uint256[][2] calldata pairs
```
cygnusv, line 883: Actually, is this what you intended? Because Solidity expresses multidimensional arrays this in the reverse order.
https://docs.soliditylang.org/en/latest/types.html#arrays

Taking this into consideration, the test below is ambiguous...

contracts/contracts/coordination/Coordinator.sol
```diff
@@ -876,4 +877,30 @@ contract Coordinator is Initializable, AccessControlDefaultAdminRulesUpgradeable
...
+        Participant memory participant = participant1;
+        participant1.provider = participant2.provider;
+        participant1.aggregated = participant2.aggregated;
+        participant1.transcript = participant2.transcript;
+        participant1.decryptionRequestStaticKey = participant2.decryptionRequestStaticKey;
+
+        participant2.provider = participant.provider;
+        participant2.aggregated = participant.aggregated;
+        participant2.transcript = participant.transcript;
+        participant2.decryptionRequestStaticKey = participant.decryptionRequestStaticKey;
```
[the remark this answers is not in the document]
Muhammad-Altabba: Resolve as this suggestion cannot be applied with solidity storage variables.

PR #433, opened 2025-10-01
Support experimental versions on evaluate_heartbeat scriptepic

manumonti:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
- Support for experimental versions. Only truly outdated versions continue to be flagged and penalized
- New `--skip-5th-heartbeat-check` CLI flag allows evaluation to run even on the 5th heartbeat of the month. Useful for edge cases requiring manual intervention
- Script now displays the list of valid versions at the start of evaluation

**Issues fixed/closed:**
Addressing comment in https://github.com/nucypher/nucypher-contracts/pull/424#discussion_r2379452265

### DIFF
--- a/scripts/evaluate_heartbeat.py
+++ b/scripts/evaluate_heartbeat.py
@@ -274,8 +274,8 @@ def format_discord_message(
     default=HEARTBEAT_ARTIFACT_FILENAME,
 )
 @click.option(
-    "--skip-5th-heartbeat-check",
-    "-s",
+    "--include-5th-heartbeat",
+    "-i",
     help="Run the evaluation even if it's the 5th heartbeat of the month.",
     is_flag=True,
     default=False,
@@ -286,9 +286,7 @@ def format_discord_message(
     is_flag=True,
     default=False,
 )
-def cli(
-    domain: str, artifact: Any, skip_5th_heartbeat_check: bool, report_infractions: bool
-) -> None:
+def cli(domain: str, artifact: Any, include_5th_heartbeat: bool, report_infractions: bool) -> None:
     """
     Evaluates the heartbeat artifact and analyzes offenders.
     This script is intended to be run shortly after a DKG heartbeat timeout to
@@ -320,7 +318,7 @@ def cli(
     offenders: Dict[str, Dict[str, Any]] = defaultdict(dict)
 
     heartbeat_round, month_name = get_heartbeat_round_info(coordinator, artifact_data)
-    if heartbeat_round > 4 and not skip_5th_heartbeat_check:
+    if heartbeat_round > 4 and not include_5th_heartbeat:
         click.secho(
             f"⚠️ This is the heartbeat round #{heartbeat_round}, which exceeds"
             + " the expected maximum of 4 per month.",


vzotova (approved):

scripts/evaluate_heartbeat.py
```diff
@@ -309,7 +319,8 @@ def cli(domain: str, artifact: Any, report_infractions: bool) -> None:
 
     offenders: Dict[str, Dict[str, Any]] = defaultdict(dict)
 
-    if heartbeat_round > 4:
+    heartbeat_round, month_name = get_heartbeat_round_info(coordinator, artifact_data)
+    if not skip_5th_heartbeat_check and heartbeat_round > 4:
```
derekpierre, line 323: Did we get this logic backward?

```suggestion
    if skip_5th_heartbeat_check and heartbeat_round > 4:
```

it might be more clear as:


```suggestion
    if heartbeat_round > 4 and skip_5th_heartbeat_check:
```

scripts/evaluate_heartbeat.py
```diff
@@ -273,13 +273,22 @@ def get_ordinal_suffix(n: int) -> str:
     type=click.File("r"),
     default=HEARTBEAT_ARTIFACT_FILENAME,
 )
+@click.option(
+    "--skip-5th-heartbeat-check",
+    "-s",
+    help="Run the evaluation even if it's the 5th heartbeat of the month.",
+    is_flag=True,
+    default=False,
```
derekpierre, line 281: Is this the default we want i.e. don't skip the 5th heartbeat check by default?

i.e. based on the help did you mean `--run-5th-heartbeat-check` and then default it to false?

Or am I misreading it?

scripts/evaluate_heartbeat.py
```diff
@@ -354,8 +365,14 @@ def cli(domain: str, artifact: Any, report_infractions: bool) -> None:
                 try:
                     version = Version(version)
                     if version not in valid_versions:
-                        offenders[address]["reasons"].append(OUTDATED)
-                        click.secho(f"Node {address} is outdated: {version}", fg="cyan")
+                        if version > valid_versions[0]:
+                            click.secho(
+                                f"⚠️ Node {address} is running an experimental version: {version}",
```
derekpierre, line 370: 👍

scripts/evaluate_heartbeat.py
manumonti: I think I'm not using the most appropriate wording and it's causing confusion.

## What I would like to have is:

- By default, the 5th heartbeat check is enabled. So, if the CI run this script on the 5th Monday of the month, the script will be skipped. That's the reason of having `--skip-5th-heartbeat-check` flag to `default=False`: I don't want to skip this check by default.
- But I want to have the possibility of running the script manually sometimes (like for the last heartbeat Sep29, that is the 5th Monday of September). So we have the flag `--skip-5h-heartbeat-check`, that will skip this check and, therefore, will allow me to run the script.

So, regarding 

https://github.com/nucypher/nucypher-contracts/pull/433/files/4ca01d4d6b0fd7bef15b1154f4078f65bc13c6e8#diff-d4ff4f6b2bad467cc262066d30f13324b3b2e01bc73043a896b3899ef59d1faeR323

what I'm trying to run is:
```
if heartbeat_round > 4      # if this is the 5th heartbeat
AND
not skip_5th_heartbeat_check      # we are not skipping the check that verifies if we are in the 5th heartbeat
THEN
return # don't run the script
```

## Example
If I run the script without the `--skip-5th-heartbeat-chck` flag for the last heartbeat (again that's the 5th for September), I'll get:

```
▶ ape run evaluate_heartbeat --network polygon:mainnet:infura --domain mainnet --artifact heartbeat-rituals.json
🔍 Analyzing DKG protocol violations...
List of valid versions: 7.6.1, 7.6.0
⚠️ This is the heartbeat round #5, which exceeds the expected maximum of 4 per month.
Skipping heartbeat evaluation...
```

Since I want it to be run because I'm debugging the script, I can run:

```
▶ ape run evaluate_heartbeat --network polygon:mainnet:infura --domain mainnet --artifact heartbeat-rituals.json --skip-5th-heartbeat-check
🔍 Analyzing DKG protocol violations...
List of valid versions: 7.6.1, 7.6.0
Node 0xAAFc71044C2B832dDDFcedb0AE99695B0367dC57 didn't send transcript
Node 0x24a278d2873879dcC8a17f504A75BFfcD4A3153a didn't send transcript
[...]
```

I agree that it can be a bit confusing because I'm using negative logic. If you have some suggestion it would be great!

derekpierre: Thanks for the clarification. It sounds like:
1. You don't want the 5th heartbeat to be evaluated by default
2. You want to be able to override locally

What if we use something more affirmative and make it opt-in. What do you think about:
* Flag name: `--check-5th-heartbeat`
* Default value: `False`
* the if statement would then be 
```python
if heartbeat_round > 4 and not check_5th_heartbeat:
     return # don't run rest of script
```
* Whenever you want to locally override that setting you use the flag in the CLI.

derekpierre (approved):
🎸

PR #444, opened 2025-11-12
[WIP] Conditions Hierarchy

vzotova:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/coordination/SigningCoordinator.sol
+++ b/contracts/contracts/coordination/SigningCoordinator.sol
@@ -53,9 +53,9 @@ contract SigningCoordinator is Initializable, AccessControlDefaultAdminRulesUpgr
         uint256[20] gap;
     }
 
-    struct Server {
-        address[] serverAdmins;
-        mapping(address serverAdmin => bool active) activeServerAdmins;
+    struct Domain {
+        address[] domainAdmins;
+        mapping(address domainAdmin => bool active) activeDomainAdmins;
         mapping(uint256 chainId => bytes conditions) conditions;
     }
 
@@ -71,7 +71,7 @@ contract SigningCoordinator is Initializable, AccessControlDefaultAdminRulesUpgr
         SigningCohortParticipant[] signers;
         uint256[] chains;
         mapping(uint256 chainId => bytes conditions) conditions; // TODO: chainId -> condition itself or hash(condition)
-        mapping(uint256 serverId => Server) serverConditions;
+        mapping(uint256 domainId => Domain) domainConditions;
     }
 
     bytes32 public constant INITIATOR_ROLE = keccak256("INITIATOR_ROLE");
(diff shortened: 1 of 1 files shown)


contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -292,40 +293,62 @@ contract SigningCoordinator is Initializable, AccessControlDefaultAdminRulesUpgr
...
+        uint256 serverId
+    ) external {
+        SigningCohort storage signingCohort = signingCohorts[cohortId];
+        conditionsCheck(signingCohort, chainId);
+        require(serverId > 0, "Server id can't be zero");
+        Server storage server = signingCohort.serverConditions[serverId];
 
         require(
-            signingCohort.authority == msg.sender || serverAdmin.active,
+            signingCohort.authority == msg.sender || server.activeServerAdmins[msg.sender],
```
derekpierre: Does this mean that the cohort authority can override a server admin for setting the serverId-specific condition? If so, I don't think we want to give the cohort authority that power.

Only the server admin can set the serverId-specific condition. (cc @arjunhassard )

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -46,6 +53,12 @@ contract SigningCoordinator is Initializable, AccessControlDefaultAdminRulesUpgr
         uint256[20] gap;
     }
 
+    struct Server {
```
cygnusv: I think we should find a more generic term for `Server`, which is too specific to a particular use case. What about something like `Domain`?

contracts/contracts/coordination/SigningCoordinator.sol
cygnusv: But cohort admin can set itself to server admin anyway, right?

cygnusv (commented):
some very early comments

contracts/contracts/coordination/SigningCoordinator.sol
```diff
@@ -265,17 +295,101 @@ contract SigningCoordinator is Initializable, AccessControlDefaultAdminRulesUpgr
...
+    function setSigningCohortConditions(
+        uint32 cohortId,
+        uint256 chainId,
+        bytes calldata conditions,
+        uint256 serverId,
+        bytes calldata signature
+    ) external {
+        SigningCohort storage signingCohort = signingCohorts[cohortId];
+        conditionsCheck(signingCohort, chainId);
+        require(serverId > 0, "Server id can't be zero");
```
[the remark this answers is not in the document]
vzotova: some optimization for `getSigningCohortConditions`

contracts/contracts/coordination/SigningCoordinator.sol
vzotova: Removed privilege for cohort admin

PR #471, opened 2026-04-01
Follow-up PR for Compensation with Penalties

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

This PR extends `PenaltyBoard` with a fixed compensation model for stakers:
 - a constant amount per period in a configurable ERC20
 - withdrawable by the staking provider, owner, or beneficiary. Tokens are always sent to the **beneficiary** . 
 - Compensation is disabled for stakeless stakers.
 - Penalties in the “last 4 periods” (configurable window) reduce or zero out compensation for affected periods via a simple factor (no penalties → full; any penalty in window → none for that period). This will be tuned in later PRs
 - Compensation logic lives in the same contract as penalty data.
 - Penalty storage is **staker-centric**: `setPenalizedProvidersForPeriod` appends the period index to each provider’s list; there is no on-chain period→providers mapping (subgraph can provide that view). This is a departure from the original PR #456, where storage was period-centric.
 - The contract depends on `TACoApplication` for roles and beneficiary. 
 - Accrual algorithm is designed to be efficient and allows for accumulated withdrawals.
    - Accrual is lazy (computed on `getAccruedBalance` and persisted on `withdraw`).
    - For a staker we take the accrual window from the last settled period (or period 0 if never withdrawn) to the current period.
    - We fetch penalized periods in the range `[startPeriod - PENALTY_WINDOW_PERIODS, current]`.
    - If there are none, we add `numPeriods × fixedCompensationPerPeriod`.
    - Otherwise, for each period P in the window we check whether any penalty `k` in that range falls in `[P - window, P]`; if so, that period gets zero, else full. 

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -123,7 +123,6 @@ contract TACoApplication is ITACoChildToRoot, OwnableUpgradeable {
     uint256 public immutable minOperatorSeconds;
 
     IERC20 public immutable token;
-    PenaltyBoard public immutable penaltyBoard;
 
     ITACoRootToChild public childApplication;
     address private _adjudicator;
@@ -144,26 +143,20 @@ contract TACoApplication is ITACoChildToRoot, OwnableUpgradeable {
 
     // mapping(address => bool) public stakingProviderReleased;
     // mapping(address => bool) public allowList;
+    PenaltyBoard public penaltyBoard;
 
     /**
      * @notice Constructor sets address of token contract and parameters for staking
      * @param _token T token contract
      * @param _minimumAuthorization Amount of minimum allowable authorization
      * @param _minOperatorSeconds Min amount of seconds while an operator can't be changed
-     * @param _penaltyBoard PenaltyBoard contract
      */
-    constructor(
-        IERC20 _token,
-        uint96 _minimumAuthorization,
-        uint256 _minOperatorSeconds,
-        PenaltyBoard _penaltyBoard
-    ) {
+    constructor(IERC20 _token, uint96 _minimumAuthorization, uint256 _minOperatorSeconds) {
         uint256 totalSupply = _token.totalSupply();
         require(totalSupply > 0, "Wrong input parameters");
         minimumAuthorization = _minimumAuthorization;
         token = _token;
         minOperatorSeconds = _minOperatorSeconds;
-        penaltyBoard = _penaltyBoard;
         _disableInitializers();
     }
 
@@ -192,6 +185,14 @@ contract TACoApplication is ITACoChildToRoot, OwnableUpgradeable {
     function setChildApplication(ITACoRootToChild _childApplication) external onlyOwner {
         require(address(_childApplication).code.length > 0, "Child app must be contract");
         childApplication = _childApplication;
+    }
+
+    /**
+     * @notice Set contract for compensation
+     */
+    function setPenaltyBoard(PenaltyBoard _penaltyBoard) external onlyOwner {
+        require(address(_penaltyBoard).code.length > 0, "PenaltyBoard must be contract");
+        penaltyBoard = _penaltyBoard;
     }
 
     //------------------------Staking------------------------------
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -122,6 +122,7 @@ contract TACoApplication is ITACoChildToRoot, OwnableUpgradeable {
     uint256 public immutable minOperatorSeconds;
 
     IERC20 public immutable token;
+    // PenaltyBoard public immutable penaltyBoard;
 
     ITACoRootToChild public childApplication;
     address private _adjudicator;
--- a/contracts/contracts/TACoApplication.sol
+++ b/contracts/contracts/TACoApplication.sol
@@ -9,6 +9,7 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin-upgradeable/contracts/access/OwnableUpgradeable.sol";
 import "./coordination/ITACoRootToChild.sol";
 import "./coordination/ITACoChildToRoot.sol";
+import "./coordination/PenaltyBoard.sol";
 
 /**
  * @title TACo Application
@@ -122,7 +123,7 @@ contract TACoApplication is ITACoChildToRoot, OwnableUpgradeable {
     uint256 public immutable minOperatorSeconds;
 
     IERC20 public immutable token;
-    // PenaltyBoard public immutable penaltyBoard;
+    PenaltyBoard public immutable penaltyBoard;
 
     ITACoRootToChild public childApplication;
     address private _adjudicator;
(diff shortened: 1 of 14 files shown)


contracts/contracts/coordination/PenaltyBoard.sol
```diff
@@ -0,0 +1,303 @@
...
+        address admin,
+        address _tacoApplication,
+        address _token,
+        uint256 _fixedCompensationPerPeriod,
+        uint256 _fixedCompensationPerPeriod2Penalties,
+        uint256 _fixedCompensationPerPeriod3Penalties,
+        address _fundHolder,
+        uint256 _startPeriodForRewards
+    ) Periods(genesisTime, periodDuration) {
+        require(admin != address(0), "Admin required");
```
derekpierre, line 73: Should we also add some other require statements...?:

```suggestion
        require(admin != address(0), "Admin required");
        require(_fixedCompensationPerPeriod2Penalties <= _fixedCompensationPerPeriod, "Invalid 2 penalty compensation");
        require(_fixedCompensationPerPeriod3Penalties <= _fixedCompensationPerPeriod2Penalties, "Invalid 3 penalty compensation");
```

contracts/contracts/TACoApplication.sol
```diff
@@ -533,4 +546,9 @@ contract TACoApplication is ITACoChildToRoot, OwnableUpgradeable {
         emit StakelessProviderAdded(_stakingProvider);
         _updateAuthorization(_stakingProvider, info);
     }
+
+    function isEligibleForReward(address _stakingProvider) external view returns (bool) {
+        StakingProviderInfo storage info = stakingProviderInfo[_stakingProvider];
+        return !info.stakeless && info.operatorConfirmed;
```
derekpierre, line 574: Out of curiosity, any reason we don't check for `info.authorized > 0` here as well?

deployment/constructor_params/mainnet/infraction.yml
```diff
@@ -0,0 +1,24 @@
...
+  dir: ./deployment/artifacts/
+  filename: mainnet-infraction.json
+
+constants:
+  # See deployment/artifacts/mainnet.json
+  COORDINATOR_PROXY: "0xE74259e3dafe30bAA8700238e324b47aC98FE755"
+  # Threshold Network - References:
+  # - https://docs.threshold.network/resources/contract-addresses/mainnet/threshold-dao
+  # - https://github.com/keep-network/tbtc-v2/issues/594
+  THRESHOLD_COUNCIL_ON_POLYGON: "0x9F6e831c8F8939DC0C830C6e492e7cEf4f9C2F5f"
```
derekpierre, line 15: Should this be the NuCo Multisig instead?

scripts/tapir/deploy_infraction.py
```diff
@@ -0,0 +1,20 @@
...
+CONSTRUCTOR_PARAMS_FILEPATH = CONSTRUCTOR_PARAMS_DIR / "tapir" / "infraction.yml"
+TAPIR_REGISTRY = ARTIFACTS_DIR / "tapir.json"
+
+
+def main():
+    deployer = Deployer.from_yaml(filepath=CONSTRUCTOR_PARAMS_FILEPATH, verify=VERIFY)
+    infraction = deployer.deploy(project.InfractionCollector)
+    deployments = [infraction]
+    deployer.finalize(deployments=deployments)
+
```
derekpierre, line 20: Need to merge the registry (similar to what is done in the lynx script)?:
```suggestion
    merge_registries(
        registry_1_filepath=TAPIR_REGISTRY,
        registry_2_filepath=deployer.registry_filepath,
        output_filepath=TAPIR_REGISTRY,
    )
```

scripts/tapir/deploy_infraction.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/python3
+
+from ape import project
+
+from deployment.constants import (
+    CONSTRUCTOR_PARAMS_DIR, ARTIFACTS_DIR,
+)
+from deployment.params import Deployer
+
```
derekpierre, line 9:
```suggestion
from deployment.registry import merge_registries

```

tests/application/conftest.py
```diff
@@ -81,3 +81,10 @@ def child_application(project, creator, taco_application):
     )
     taco_application.setChildApplication(contract.address, sender=creator)
     return contract
+
+
+@pytest.fixture()
+def penalty_board(project, creator, taco_application):
+    contract = project.PenaltyBoardForTACoApplicationMock.deploy(sender=creator)
+    taco_application.setPenaltyBoard(contract.address, sender=creator)
```
derekpierre, line 89: Any reason you don't want to include the setting of the penalty board in the taco_application fixture itself? Then you wouldn't have to include the penalty_board fixture in tests where you aren't making assertions on the penalty board contract? For example, `test_request_unstake`, `test_release`, `test_child_sync`.

i.e. you can potentially do the following instead:
def taco_application(..., pentalty_board):
...
taco_application.setPenaltyBoard(contract.address, sender=creator)
def penalty_board(project, creator):
contract = project.PenaltyBoardForTACoApplicationMock.deploy(sender=creator)
return contract

tests/test_penalty_board.py
```diff
@@ -0,0 +1,143 @@
...
+
+    current = penalty_board.getCurrentPeriod()
+    # At deploy, genesis = chain.pending_timestamp so current is 0. Period 1 is invalid.
+    with ape.reverts("Invalid period"):
+        penalty_board.addPenalizedProvidersForPeriod(provs, current + 1, sender=informer)
+
+    penalty_board.addPenalizedProvidersForPeriod(provs, current, sender=informer)
+    assert penalty_board.getPenalizedPeriodsByStaker(other_account.address) == [current]
+
+    # Advance into period 1; then both 2 and 3 are valid.
```
derekpierre, line 90: Is it period 1 or 3 for the comment?

```suggestion
    # Advance into period 3; then both 2 and 3 are valid.
```

tests/test_penalty_board_compensation.py
```diff
@@ -0,0 +1,473 @@
...
+    return staking_provider
+
+
+@pytest.fixture
+def registered_stakeless(mock_taco_app, stakeless_provider, owner, beneficiary, deployer):
+    """Register stakeless_provider in mock TACo: owner=owner, beneficiary=beneficiary, stakeless."""
+    mock_taco_app.setRoles(
+        stakeless_provider.address,
+        owner.address,
+        beneficiary.address,
```
derekpierre: Can a `stakeless` node have a beneficiary?

derekpierre (commented):
Nice work. Still making my way through. Some comments in the meantime.

contracts/contracts/coordination/PenaltyBoard.sol
vzotova: well it's not breaking any math there so not really necessary. I'll add it if you would like

contracts/contracts/TACoApplication.sol
vzotova: I don't see how authorized can be zero if operator confirmed so feels overkill

deployment/constructor_params/mainnet/infraction.yml
cygnusv: This is outdated code from the epic  branch. Will be addressed in a different PR

